PR #13, opened 2019-01-30
Feature/homepage 10

### LINKED ISSUE #10: Search/Home page
Search Fields:
* Tags de área (de informática): Back-End, Front-End, DevOps, etc - Também de tecnologias mais comuns
* Field de localização (Enum por distritos - primeira ideia) -> investigar API google maps para isso
* Tipo de emprego: Full-time, Part-time, Estágio Curricular/Verão, etc

Get layout inspirations from Google Flights and Google carrers -> jobs webpages

imnotteixeira:
> Huge PR, I am sorry.

Closes #10 and #18. 

This PR sets up the bar for what I hope to be a big project, it contains a new homepage, which involves a lot of small components to allow for separation of concerns and easier testing. The components state is now managed with React Hooks (https://reactjs.org/docs/hooks-intro.html), and I hope somewhere in the future we'll migrate the redux state to hooks as well.

**Testing**. This PR's code is fully tested, as should every PR from now on. The coverage limit is >80%, even though this is not fully representative of good testing, it's a nice starting point.

**Material UI** was also updated to v4 and the way we use styles and the general theme was updated as well

A new product is emerging, let this be a motivation for new NIAEFEUP members to contribute and learn new things!


miguelpduarte:
Can this be changed into a draft PR so no accidental merging occurs?

imnotteixeira:
Unfortunately not, as it was created as a non-draft PR, or at least that's what I think. 

If you can, try to mark it as a draft yourself, but as far as I know, that is not possible in this situation.

This is a good input however, for future PR creation, as they should always be in Draft Mode while in feature discussion to avoid accidental merge.

miguelpduarte:
I think you're right, and maybe that is also because this was probably created before GitHub supported draft PRs.

Might also look into using something like [`github terraform module`](https://www.terraform.io/docs/providers/github/index.html) for configuring branch protection rules to be standardized throughout all our repos, as I'm seeing some mistmatches here and there. But let's not lose scope, this should be discussed elsewhere (NIAEFEUP/meta really 😅)

src/App.js
```diff
@@ -4,17 +4,22 @@ import AppRouter from './AppRouter';
 import store from './store';
 import Notifier from './components/Notifications/Notifier';
 
+
 // import './App.css';
```
miguelpduarte: Remove dead code?

src/App.js
```diff
@@ -4,17 +4,22 @@ import AppRouter from './AppRouter';
...
+
         return (
             <Provider store={store}>
+                
                 <div>
                     {/* <AppNavbar/> */}
                     <Notifier />
                     <AppRouter/>
                 </div>
+
```
miguelpduarte: Remove what seems like unnecessary whitespace and `div`. Do we need better eslint rules?
Can work on that if it is necessary.

src/components/HomePage/MainView.js
```diff
@@ -0,0 +1,47 @@
...
+import InfoBox from "./QuickInfoArea/InfoBox";
+
+import homePageStyles from './HomePage.module.css';
+
+const MainView = props => {
+
+    const {
+        scrollToProductDescription, 
+        scrollToSearchResults
+    } = props;
```
miguelpduarte: Use object destructuring for `props`

src/components/HomePage/MainView.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import logo from './nijobs.png';
+
+import SearchArea from "./SearchArea/SearchArea";
+import ShowMoreButton from "./ShowMoreButton";
+import InfoBox from "./QuickInfoArea/InfoBox";
+
+import homePageStyles from './HomePage.module.css';
```
miguelpduarte: Debate: Using CSS Modules vs Using the new MUI styles hooks, which would be best?

src/components/HomePage/QuickInfoArea/InfoBox.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import homePageStyles from '../HomePage.module.css';
+
+import Typography from '@material-ui/core/Typography';
+
+class InfoBox extends Component {
```
miguelpduarte: Convert to functional component to reduce boilerplate and use object destructuring to get the props

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -0,0 +1,170 @@
...
+function SearchArea(props) {
+
+    const {addSnackbar, onSubmit} = props;
+    const classes = useStyles();
+    
+    // Set initial form values
+    const [searchValue, setSearchValue] = React.useState("");
+    const [advancedSearch, setAdvancedSearch] = React.useState(false);
+    const [jobType, setJobType] = React.useState(INITIAL_JOB_TYPE);
+    const [jobDuration, setJobDuration] = React.useState(INITIAL_JOB_DURATION);
```
miguelpduarte: Import the hook directly? Personal preference, tbh, but would rather always use the hooks without the trailing `React` - will also look better when mixing with custom hooks or hooks from other sources (I spy one just above, `useStyles`, that looks out of place due to this)

src/components/HomePage/MainView.js
```diff
@@ -0,0 +1,47 @@
...
+            </div>
+            <InfoBox
+                info='Your next oportunity is out there. Use the search bar to find it!'
+            />
+            <SearchArea onSubmit={scrollToSearchResults}/>
+            <ShowMoreButton 
+                className={homePageStyles.showMoreBtn} 
+                onClick={scrollToProductDescription}
+            />
+
```
miguelpduarte: whitespace

src/AppTheme.js
```diff
@@ -0,0 +1,33 @@
...
+        fontFamily: [
+            'Poppins',
+            'Roboto',
+            'sans-serif',
+        ].join(','),
+        useNextVariants: true,
+    },
+
+}); 
+export default theme;
```
miguelpduarte: Are we using a `ThemeProvider`? Might be better to not have to explicitly include the theme everywhere, just wrap some bottom level component, near `App` or something.

src/AppRouter.js
```diff
@@ -1,11 +1,11 @@
...
+import HomePageBoilerplate from './pages/HomePageBoilerplate';
 import HomePage from './pages/HomePage';
 import ExamplePage from './pages/ExamplePage';
 import OtherPage from './pages/OtherPage';
 import NotFound from './pages/NotFound';
 
-
 class AppRouter extends Component {
     render(){
         return (
```
miguelpduarte: Is the React.Fragment still necessary or left-over?

src/App.js
```diff
@@ -4,17 +4,22 @@ import AppRouter from './AppRouter';
...
 
 class App extends Component {
+
     render() {
+
         return (
             <Provider store={store}>
+                
                 <div>
                     {/* <AppNavbar/> */}
```
miguelpduarte: Remove dead code?

src/components/HomePage/ProductPlacementArea/ProductDescription.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import productDescriptionStyle from './ProductDescription.module.css';
+
+const ProductDescription = props => {
+
+    const {setRef} = props;
+    const ref = React.useRef(null);
+    setRef(ref);
```
miguelpduarte: Is this because of scrolling? Would it be better to use an id directly or is there some limitation in that approach?

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -0,0 +1,170 @@
...
+    addSnackbar: PropTypes.func,
+    onSubmit: PropTypes.func.isRequired
+};
+
+const mapStateToProps = () => ({
+});
+
+const mapActionsToProps = {addSnackbar};
+
+export default connect(mapStateToProps, mapActionsToProps)(SearchArea);
```
miguelpduarte: use hooks for consistency?

src/components/HomePage/SearchArea/SearchBar.js
```diff
@@ -0,0 +1,54 @@
...
+        <TextField
+            label="Search"
+            name="search"
+            margin="dense"
+            value={searchValue}
+            onChange={handleChange}
+            className={className}
+            InputProps={{
+                "endAdornment": 
+    <InputAdornment position="end">
```
miguelpduarte: Uh oh, eslint spaghetti?

miguelpduarte: Looks like weird indentation now, because of being nested and a value of a key... Is there a better way to pass that prop? This looks ugly

src/components/HomePage/SearchArea/SearchBarTheme.js
```diff
@@ -0,0 +1,32 @@
...
+    },
+    notchedOutline: {},
+    formControl: {},
+    
+
+    searchButton: {
+        color: darkTextColor,
+    },
+});
+
```
miguelpduarte: Themes inside themes? Can't this all be set in the global theme and passed to everyone that wants it using a `ThemeProvider`? Would simplify things as then it would be as simple as using `makeStyles` or `useTheme` for everything, iirc 😃

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -0,0 +1,170 @@
...
+    const {addSnackbar, onSubmit} = props;
+    const classes = useStyles();
+    
+    // Set initial form values
+    const [searchValue, setSearchValue] = React.useState("");
+    const [advancedSearch, setAdvancedSearch] = React.useState(false);
+    const [jobType, setJobType] = React.useState(INITIAL_JOB_TYPE);
+    const [jobDuration, setJobDuration] = React.useState(INITIAL_JOB_DURATION);
+    
+    const toggleAdvancedOptions = () => {
```
miguelpduarte: This seems extractable to a `useToggledState` or similarly-named hook, exporting something like `[state, toggleState]`

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -0,0 +1,170 @@
...
+                    vertical: 'top',
+                    horizontal: 'right',
+                },
+            }
+        });
+
+        onSubmit();
+    };
+
+    const updateSearchValue = newVal => {
```
miguelpduarte: is it not possible to pass `setSearchValue` directly? Would reduce code clutter

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -0,0 +1,170 @@
...
+        setSearchValue(newVal);
+    };
+    
+    const updateJobDuration = (_, val) => {
```
miguelpduarte: Can `setJobDuration` be passed using some `.bind` shenanigans? Would reduce code clutter

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
```diff
@@ -0,0 +1,30 @@
...
+
+const ShowAdvancedOptionsButton = ({open, onClick}) => {
+    return (
+        <div className={searchAreaStyle.advancedSearchBtnWrapper}>
+            <Fab
+                color="primary"
+                aria-label="Show More Options"
+                onClick={onClick}
+            >
+                {open ? <Close /> : <MoreHoriz fontSize="large"/>}
```
miguelpduarte: Would kinda expect a Close/Open pair but can understand the usage of MoreHoriz. Could/Should be debated

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Fab from '@material-ui/core/Fab';
+import Close from '@material-ui/icons/CloseRounded';
+import MoreHoriz from '@material-ui/icons/MoreHoriz';
```
miguelpduarte: Use tree-shaking for imports (I believe our configs allow this) - it's suggested in the MUI docs.

miguelpduarte: This but everywhere, just noticed it

src/components/HomePage/SearchArea/SliderValueTooltip.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Tooltip from '@material-ui/core/Tooltip';
+
+const SliderValueTooltip = props => {
+    const { children, open, value } = props;
```
miguelpduarte: Object destructuring for prop getting

src/components/HomePage/SearchArea/jobTypes.js
```diff
@@ -0,0 +1,25 @@
...
+        label: "Part-Time"
+    },
+    {
+        value: "FULL_TIME",
+        label: "Full-Time"
+    },
+
+]);
+
+export default JOB_TYPES;
```
miguelpduarte: Should this be a request to the backend? Can also understand hardcoding, as this will likely never change, but would like for consistency :/

src/components/HomePage/SearchResultsArea/Offer/OfferCard.js
```diff
@@ -0,0 +1,86 @@
...
+
+import PropTypes from 'prop-types';
+
+import { makeStyles } from '@material-ui/core/styles';
+import Card from '@material-ui/core/Card';
+import CardHeader from '@material-ui/core/CardHeader';
+//import CardActions from '@material-ui/core/CardActions';
+import CardContent from '@material-ui/core/CardContent';
+import CardMedia from '@material-ui/core/CardMedia';
+//import Button from '@material-ui/core/Button';
```
miguelpduarte: Remove dead code and switch to using tree-shaking

src/components/HomePage/SearchResultsArea/Offer/OfferCard.js
```diff
@@ -0,0 +1,86 @@
...
+
+import OfferCardTheme from './OfferCardTheme';
+
+const logo = require('./ni.png');
+
+const useStyles = makeStyles(OfferCardTheme);
+
+const OfferCard = props => {
+
+    const { loading, position, company, location, date, description } = props;
```
miguelpduarte: object destructuring for props

src/components/HomePage/SearchResultsArea/Offer/OfferCardTheme.js
```diff
@@ -0,0 +1,39 @@
+import AppTheme from '../../../../AppTheme';
+
```
miguelpduarte: Same concerns as with `SearchBarTheme`, seems weird... Would probably prefer this be done directly in the component, otherwise it might get confusing (despite the added code clutter)

src/components/HomePage/ShowMoreButton.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import KeyboardArrowDownIcon from '@material-ui/icons/KeyboardArrowDownRounded';
+import IconButton from '@material-ui/core/IconButton';
+
+const ShowMoreButton = props => {
+    
+    const { className, onClick } = props;
```
miguelpduarte: you know it - object destructuring!

src/components/Notifications/Notifier.js
```diff
@@ -5,36 +5,40 @@ import { withSnackbar } from 'notistack';
...
 
-    render() {
-        const { notifications, enqueueSnackbar, removeSnackbar } = this.props;
-        const { displayed } = this.state;
+    shouldComponentUpdate({ notifications: newSnacks = [] }) {
+        const { notifications: currentSnacks } = this.props;
+        let notExists = false;
+        for (let i = 0; i < newSnacks.length; i += 1) {
+            if (notExists) continue;
+            notExists = notExists || !currentSnacks.filter(({ key }) => newSnacks[i].key === key).length;
```
miguelpduarte: Looks like a candidate for optimization, returning as soon as the condition evaluates to true, and returning false outside the loop.

Would reduce the number of iterations and remove one variable

EDIT: Just noticed the `continue` - makes me feel that this refactor would be guaranteed to be less smelly code! 😄

src/index.js
```diff
@@ -3,12 +3,17 @@ import ReactDOM from 'react-dom';
...
+
 import './index.css';
 import App from './App';
 
 ReactDOM.render(
     <BrowserRouter>
-        <SnackbarProvider maxSnack={3}>
-            <App />
-        </SnackbarProvider>
+        <ThemeProvider theme={AppTheme}>
```
miguelpduarte: Disregard the comments on a `ThemeProvider` made previously 🤦‍♂

src/pages/HomePage.js
```diff
@@ -1,21 +1,38 @@
...
-        return (
-            <React.Fragment>
-                <Banner />
-                <ButtonBar />
-                <SleepyActionOutput />
-                <RandomDogOutput />
-            </React.Fragment>
-        );
-    }
+function HomePage() {
```
miguelpduarte: Arrow functions? Yes please!

src/pages/HomePage.js
```diff
@@ -1,21 +1,38 @@
...
+        searchResultsRef.current.scrollIntoView({
+            behavior: "smooth",
+            block: "start"
+        });
+    };
+
+    return (
+        <React.Fragment>
+            <MainView 
+                scrollToProductDescription={scrollToProductDescription}
```
miguelpduarte: These seem like they are being passed on for several levels. This is less than ideal, but I do not feel very well about putting them in the redux store either... :/

Would a solution using html `id`s work? It might be better code-wise

src/pages/HomePageBoilerplate.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
+
```
miguelpduarte: I'd vote in favor of removing this, as with all of the example files

src/test/App.test.js
```diff
@@ -1,8 +1,15 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from '../App';
+import { BrowserRouter } from "react-router-dom";
+import { SnackbarProvider } from 'notistack';
```
miguelpduarte: Maybe we can look into snapshot testing (I think this is simple-ish to setup using `jest` iirc) as a way to kickstart initial testing practices, and then develop into better tests, eventually using something like [`react-testing-library`](https://github.com/testing-library/react-testing-library) - I've used it a little, and it's pretty cool.

src/components/HomePage/ProductPlacementArea/ProductDescription.module.css
```diff
@@ -0,0 +1,4 @@
+.container {
+    /* todo:remove (this height is to simulate placeholder) */
```
miguelpduarte: Comment as a reminder to remove comments! 😆

src/App.js
imnotteixeira: Yes, you can work on that if you want. While you're at it, try to add react hooks eslint rules as well

miguelpduarte: When I did it was as simple as adding a package. Will investigate some options and try working on it tomorrow night or this week 👍

src/AppTheme.js
imnotteixeira: We are doing that

src/App.js
miguelpduarte: (For reference: https://www.npmjs.com/package/eslint-plugin-react-hooks)

src/components/HomePage/MainView.js
imnotteixeira: I think both have their place. Currently I am using the MUI hooks for overrides in the MUI components themselves, while the CSS Modules serve as general CSS rules for the application (div placement, grids, etc)

src/AppTheme.js
miguelpduarte: Yeah I found that out in the end, thank the file ordering 🙃  🤦‍♂ 

Ignore this and other comments relative to this, just resolve them straight away

src/components/HomePage/MainView.js
imnotteixeira: I thought about it when making this, but I figured it would be more readable if it was like this, in case the props increase.

src/components/HomePage/ProductPlacementArea/ProductDescription.js
imnotteixeira: Refs must be used to access the html element the React way, so this is the right way

src/components/HomePage/ProductPlacementArea/ProductDescription.module.css
imnotteixeira: Well, that whole container is still implementation pending, so I believe the comment will be dealt with xD

src/components/HomePage/QuickInfoArea/InfoBox.js
imnotteixeira: Currently in the process of migrating existing components into functional ones

src/components/HomePage/SearchArea/SearchArea.js
imnotteixeira: I think we should use `React.` for the vanilla hooks, and for the custom ones, such as `useStyles()`, omit it of course

src/components/HomePage/MainView.js
miguelpduarte: Makes sense, this way the CSS Modules maintain their broader scope and our CSS does not end up like 🍝, while taking advantage of MUI overrides 😄 

Still, might be better to document these decisions somewhere, I'm thinking throwing something on the README for now would be good - what do you think? We should track these as issues or somewhere else too, if we don't do them straight away (don't have a Jira, PivotalTracker or simillar for now so GitHub Issues will do).

Should also look into starting some less specific CSS Modules - I'm certain some placement, sizing, or coloring ones will eventually be useful, and this would improve modularity - a concern to keep in mind for the future.

src/components/HomePage/SearchArea/SearchArea.js
imnotteixeira: This PR is undergoing a transition into the new Hooks API + Functional Components. I believe all the components logic could be improved, so do not let me forget that suggestion

src/components/HomePage/SearchArea/SearchArea.js
imnotteixeira: It is possible. !remindme

src/components/HomePage/MainView.js
miguelpduarte: Unlikely imo, but can understand the idea. I personally prefer to go with the flow, as a little bit of future proofing can easily become too much 😅 Thus, would still vouch for destructuring 😄

src/components/HomePage/QuickInfoArea/InfoBox.js
miguelpduarte: :shipit:

src/components/HomePage/SearchArea/SearchArea.js
imnotteixeira: That time will come someday... :P

src/components/HomePage/ProductPlacementArea/ProductDescription.js
miguelpduarte: Fair point, did not know about that! A quick search prooved me wrong, though 🤦‍♂ 
Truck away and ignore some other comments regarding this, except the one debating the multi-level prop passing (redux as another possibility?)

src/components/HomePage/ProductPlacementArea/ProductDescription.module.css
miguelpduarte: Glad to hear it! Nuke all the comments! 🚀  💥

src/components/HomePage/SearchArea/SearchBar.js
imnotteixeira: Definitely eslint spaghett, at the rules definition time, we could not find a reliable way of effectively manage the components props, maybe the situation has changed now. Needs to be reviewed.

src/components/HomePage/SearchArea/SearchBarTheme.js
imnotteixeira: This theme is specific to a component, that way it should not be in a global place

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
imnotteixeira: Tree shaking must be configured at webpack level, since we created the project with create-react-app, I am not sure if it is possible to change now

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
imnotteixeira: The thoughts were the following:
- There is already a chevron to see more of the page
- Since the inputs are reset when the advanced are is close, the user should know the action is to "cancel" rather than closing the section, hence the cross instead of the previous chevron-up

src/components/HomePage/SearchArea/jobTypes.js
imnotteixeira: Probably should be moved to BE in the future.

src/components/HomePage/SearchArea/SearchArea.js
miguelpduarte: Simply put, really don't see any point for wasting characters writing `React.` everywhere :/
Can see the point you make for it, but think that the name and import already distinguishes the hook as a vanilla one (if someone tries to create custom hooks with the same name as any of the vanilla ones, I will personally nuke their PRs, _no problemo_).

This would still make hooks recognisable as `^use.*` (`useStuff`, `useSomething`), and thus the linting from [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) would still apply so don't really see any point in keeping this 😕

src/components/HomePage/SearchArea/SearchArea.js
miguelpduarte: The transition can be a bit more gradual and worked on later, if need be - I'm fine with that, the React docs themselves say that that's ok.
As for a reminder, you can just leave this conversation open, but I'd say something in the final code review too, probably 😜

src/components/HomePage/SearchResultsArea/Offer/OfferCardTheme.js
imnotteixeira: The styles and code inside the same file is a lot confusing and unreadable. Hence the separation in two files. Due to the generation of more files, the folder separation was done, which I think is a good balance (The style file naming is consistent to avoid confusion)

src/components/HomePage/SearchArea/SearchArea.js
miguelpduarte: Slack and GitHub integration incoming? Haha

src/components/Notifications/Notifier.js
imnotteixeira: This code came straight from the package docs, as I am not sure how the system works internally, maybe it is intended to loop through all the snacks

src/index.js
imnotteixeira: :P

src/pages/HomePage.js
imnotteixeira: WIP

src/components/HomePage/SearchArea/SearchArea.js
miguelpduarte: /remind @imnotteixeira in a couple months :V

src/pages/HomePage.js
imnotteixeira: As explained above, in React, the right way to interact with the html elements is via refs, instead of using plain JS

src/pages/HomePageBoilerplate.js
imnotteixeira: Probably, but maybe in another PR

src/test/App.test.js
imnotteixeira: Agreed

imnotteixeira: As an encouragement measure, we could add code coverage rules

src/components/HomePage/SearchArea/SearchBar.js
miguelpduarte: Tracked here: https://github.com/NIAEFEUP/nijobs-fe/issues/16

src/App.js
miguelpduarte: Tracked here: https://github.com/NIAEFEUP/nijobs-fe/issues/16

src/components/HomePage/MainView.js
imnotteixeira: I think we should stick to the sizing/coloring conventions from MUI Theme directly, and in the CSS Modules only change higher-level visuals (such as alignment, for example)

src/AppRouter.js
imnotteixeira: I think it can be removed

src/components/HomePage/SearchArea/SearchBarTheme.js
miguelpduarte: Misread this, this is the argument of `makeStyles` for the `SearchBar` it seems.
As I've said further down somewhere, I'm in favor of keeping this in the component to not lose scope, despite possible increased code clutter.

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
miguelpduarte: And isn't it enabled already? Can you test including something in the "regular" way? I have as reference a gatsby project with which it worked straight away - not much but something

Also, having our own webpack configs also looks like something desireable, as I believe that some useful configs are facilitated when using it directly, i.e after ejecting from `create-react-app` (such as custom stuff for custom file extensions, might be useful for translations when that happens in the far, far, far future).

src/components/HomePage/SearchArea/SearchBarTheme.js
imnotteixeira: These styles can get quite big for trickier customizations, thus we should keep them separated to avoid doing all the work later when some component requires 100 lines of styling and upon separating it, we realize it becomes inconsistent

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
miguelpduarte: Last time I checked it was a chevron-up, hence the comment, my bad for not checking! Agree, 🚢 📦  -> :shipit: 
Resolving

src/components/HomePage/SearchArea/jobTypes.js
miguelpduarte: Maybe have both, for safety? Not sure how fast the BE will move so might be best to keep some logic in the FE for now (I've been more adverse to this in the past than I am now)

src/components/HomePage/SearchResultsArea/Offer/OfferCardTheme.js
miguelpduarte: Let's roll with it for now and see how it evolves in the future, for now I'm okay with that

My concern was more on smaller components that would just generate way too many files for what it's worth and might as well just have it inline. As said above, time will tell the best approach

src/components/Notifications/Notifier.js
miguelpduarte: The `continue` makes me think otherwise, though! I'd vote for changing it and seeing if it breaks something 🛠

src/pages/HomePage.js
miguelpduarte: Still, what about the redux store idea? Would be good to remove the multi-level prop passing - but is still kind of weird to use redux store for that...

Still, I'd be on board with that if you agreed, as I *reeeally* don't like passing props around so much

src/pages/HomePageBoilerplate.js
miguelpduarte: Fair point, new issue?

miguelpduarte: Tracked here: https://github.com/NIAEFEUP/nijobs-fe/issues/17

src/test/App.test.js
miguelpduarte: > As an encouragement measure, we could add code coverage rules

Not ideal, tbh (as code coverage is not the full picture), but would make for some good incentive early on, I agree. Will track this

miguelpduarte: Tracked: https://github.com/NIAEFEUP/nijobs-fe/issues/18

src/components/HomePage/MainView.js
miguelpduarte: Fair point, was just listing some examples off the top of my head, but agreed

src/AppRouter.js
miguelpduarte: Blow it up! 💣💥

src/components/HomePage/SearchArea/SearchBarTheme.js
miguelpduarte: I'm okay with this, will keep this in mind and see the direction to take in the future. Will need to keep an eye for file clutter for smaller components basically.

src/App.js
miguelpduarte: PR dropped: #22

src/components/HomePage/SearchArea/SearchArea.js
imnotteixeira: In the beginning I didn't see why, but that can't be done because the `Slider` `onChange` callback receives 2 arguments from the component. If it was the other way around, (we had a function with 2 args and the component only filled the second), we could bind the passed function with the first arg pre-filled

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
imnotteixeira: After doing some quick research, I noticed create-react-app already supports tree-shaking without having to eject and change webpack configs. I will look into how to refactor the code to use this feature

src/pages/HomePage.js
imnotteixeira: I think it will stay like this, because:
1. It's only 2 levels of propagation
1. I don't think it makes much sense to use redux store for functionality related callbacks, as it should be used mostly for common data storage.
1. This is a really specific set of functions that does not relate with other parts of the application, even for future components, thus it should not be in the redux store

src/components/HomePage/SearchArea/SearchArea.js
miguelpduarte: Makes perfect sense! :shipit:

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
miguelpduarte: I can help by pointing out stuff. But generally, it will be something like changing

```js
import Close from '@material-ui/icons/CloseRounded';
import MoreHoriz from '@material-ui/icons/MoreHoriz';
```

to

```js
import { Close, MoreHoriz } from '@material-ui/icons';
```

You can probably use `grep` or `vscode` to search in the whole project for this - possible candidates are `@material-ui/core` and `@material-ui/icons`, also maybe `@material-ui/styles` if it's being used somewhere.
I'm thinking some of the `eslint` rules added in #22 might help with this, namely ones that add checks for duplicated imports and such

src/pages/HomePage.js
miguelpduarte: > 1. It's only 2 levels of propagation

"only two" evolve into "only three" into "only four" and so on... It's best to tackle the problem right as it arises, and cut close to the root. I see your point though, two levels is not that bad, despite being less than ideal.

> 2. I don't think it makes much sense to use redux store for functionality related callbacks, as it should be used mostly for common data storage.

The redux store can really be used for whatever we want tbh. If this made sense then it sure would have its place there

> 3. This is a really specific set of functions that does not relate with other parts of the application, even for future components, thus it should not be in the redux store

On that we can agree. This makes sense.

I tried also investigating some options detailed [here](https://reactjs.org/docs/context.html), but I think it's probably overkill for this case, so I'd vote for keeping it as it is for now as well, and see how it evolves in the future (don't think this component will see much reuse anyway, as it's kinda specific).

src/components/Notifications/Notifier.js
imnotteixeira: I refactored it into 
```js
for (let i = 0; i < newSnacks.length; i += 1) {
   if (!currentSnacks.filter(({ key }) => newSnacks[i].key === key).length) 
        return true;
}
return false;
```

and the notifications seem to be working normally, I even checked the redux store and it also seems fine, so I think this was a nice catch

miguelpduarte: Nice! Better code is always cool!
Well, but not the best there is... The best idea is just to have [nocode](https://github.com/kelseyhightower/nocode). 😉

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
imnotteixeira: It is already done, Regex is <3. I am assuming all tree-shaking configuration is already provided by create-react-app boilerplate, as I've read on the internet, and if it on the internet, it must be true, so we are good.

src/components/HomePage/SearchArea/SearchArea.js
imnotteixeira: After looking at the Redux hooks docs (https://react-redux.js.org/next/api/hooks) it seems really simple to do the transition, however, they mention some problems that may arise in some edge cases, so I am deferring this refactor to when this solution matures

src/components/HomePage/SearchArea/SearchArea.js
imnotteixeira: Created a custom hook `useToggle` that returns `[state, toggleState]` and hides the `state = !state` logic, so the developer only needs to call `toggleState()` to update the value

src/components/HomePage/SearchArea/ShowAdvancedOptionsButton.js
miguelpduarte: If it still works after the code change on your machine (and Netlify's PR preview) it's probably fine

:shipit:

src/components/HomePage/SearchArea/SearchArea.js
miguelpduarte: Fair point. Would it make sense to track this in an issue? (Anotha' one)
Probably, I'm guessing

src/components/HomePage/SearchArea/SearchArea.js
miguelpduarte: Sounds perfect! Looking forward to reviewing everything again after a while! 🙃 😄

src/hooks/useToggle.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Custom hook for state variables that have a special boolean state only
+ * @param {*} initialValue - initial boolean value of the state variable
+ */
+const useToggle = initialValue => {
+    const [state, setState] = useState(initialValue);
+
+    const toggleState = () => {
+        setState(!state);
```
miguelpduarte: I think it will work all the same, but would it not be better to use [functional updates](https://reactjs.org/docs/hooks-reference.html#functional-updates)?

miguelpduarte:
I don't like that the bot comments here, can we have it only as a check? Is that configurable?
If this is because it's an alert and not an error, just make all the alerts errors tbh

src/hooks/useToggle.js
imnotteixeira: I don't see how that would work here, can you post the code for this case?

miguelpduarte:
```suggestion
        setState(state => !state);
```

imnotteixeira: What is the advantage of that / Disadvantage of the existing solution?

miguelpduarte: Also, kinda feel like we should use `useCallback` here

miguelpduarte: Or something similar, to avoid creating the `toggleState` function repeatedly, which might trigger re-renders

miguelpduarte: Its only dependency is `setState`, which is guaranteed (as per the [hooks api docs - see note for `useState`](https://reactjs.org/docs/hooks-reference.html#usestate)) to not change.

imnotteixeira: Now that's a good thing to have learned! Btw, the hook does not need dependencies in this case, the callback does not change, because it depends strictly from the argument, `state` in this case, so it looks like this:
```js
const toggleState = useCallback(() => {
        setState((state) => !state);
    }, []);
```

By doing this, not only we optimize performace, but we also fix a problem, because currently, if there is more than 1 variable created with `useToggle`, there would be infinite re-renders, which is solved with this `useCallback` solution.

kudos for this catch!

miguelpduarte: That was exactly my point, that's it!

I got that from the [`react-redux` docs on migrating to hooks](https://react-redux.js.org/next/api/hooks). They are pretty helpful when considering some of the caveats of hooks, because of functions being recreated since they are being declared in `render()`.
It's a good read  for both the React and the Redux side of things, highly recommend it 🙂

package.json
```diff
@@ -6,8 +6,18 @@
...
+    "react-redux": "^5.1.1",
+    "react-router-dom": "^4.3.1",
+    "react-scripts": "^3.0.1",
+    "react-select": "^3.0.4",
+    "react-transition-group": "^1.2.1",
     "@babel/core": "7.5.5",
-    "@material-ui/core": "^3.6.0",
     "@svgr/webpack": "4.3.2",
     "@typescript-eslint/eslint-plugin": "1.13.0",
     "@typescript-eslint/parser": "1.13.0",
```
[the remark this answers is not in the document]
imnotteixeira: Side effects from `eject`

public/index.html
```diff
@@ -19,7 +23,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>React App</title>
+    <title>NIJobs</title>
```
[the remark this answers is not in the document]
imnotteixeira: Waiting for meeting with UI/UX

src/App.js
```diff
@@ -1,23 +1,18 @@
...
 import { Provider } from "react-redux";
 import AppRouter from "./AppRouter";
 import store from "./store";
 import Notifier from "./components/Notifications/Notifier";
 
-// import './App.css';
+const App = () => (
+    <Provider store={store}>
+        <React.Fragment>
+            {/* <AppNavbar/> */}
```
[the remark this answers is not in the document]
imnotteixeira: The idea was to have a Navbar in the future, but as the visual design planning phase was a bit sloppy, nothing is decided yet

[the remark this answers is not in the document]
imnotteixeira: The Css Baseline suggestion is good, will add it in the future commit

src/App.js
```diff
@@ -1,23 +1,18 @@
...
+import React from "react";
 import { Provider } from "react-redux";
 import AppRouter from "./AppRouter";
 import store from "./store";
 import Notifier from "./components/Notifications/Notifier";
 
-// import './App.css';
+const App = () => (
+    <Provider store={store}>
+        <React.Fragment>
```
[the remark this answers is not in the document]
imnotteixeira: Nope, the `Provider` expects only a single `ReactElement`: https://react-redux.js.org/api/provider

src/AppRouter.js
```diff
@@ -1,38 +1,37 @@
...
+        <Switch>
+            <Route
+                exact
+                path="/"
+                component={HomePage}
+            />
+            <Route
+                exact
+                path="/demo"
+                component={HomePageBoilerplate}
```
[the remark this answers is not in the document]
imnotteixeira: Sure, in another PR, which should close #17

src/actions/notificationActions.spec.js
```diff
@@ -0,0 +1,55 @@
...
+import { notificationTypes } from "./types";
+
+describe("Notification actions", () => {
+    it("should return addSnackbar action if notification has message", () => {
+        const RANDOM_VALUE = 0.5;
+        const DATE_NOW = 1;
+
+        const mockMath = Object.create(global.Math);
+        mockMath.random = () => RANDOM_VALUE;
+        global.Math = mockMath;
```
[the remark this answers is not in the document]
imnotteixeira: You are right, fixed in next commit

src/actions/notificationActions.js
```diff
@@ -1,18 +1,21 @@
...
 
-export const removeSnackbar = (key) => (dispatch) => {
-    dispatch({
-        type: notificationTypes.REMOVE_SNACKBAR,
-        key,
-    });
-};
+export const removeSnackbar = (key) => ({
+    type: notificationTypes.REMOVE_SNACKBAR,
+    key,
```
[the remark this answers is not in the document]
imnotteixeira: There is no standard regarding Redux Actions structure (https://redux.js.org/basics/actions), check their example. As long as the reducer matches the object fields, it should be fine. The only action field that is "mandatory" is `type`

src/actions/notificationActions.js
```diff
@@ -1,18 +1,21 @@
...
+
+    if (typeof notification.message !== "string" || notification.message.length === 0) {
+        throw new Error("Notification must have a message field");
+    }
+
+    return {
         type: notificationTypes.ADD_SNACKBAR,
         notification: {
-            key: new Date().getTime() + Math.random(),
+            key: Date.now() + Math.random(),
```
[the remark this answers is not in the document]
imnotteixeira: Answered above

src/components/HomePage/MainView.js
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import logo from "./nijobs.png";
```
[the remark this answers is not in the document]
imnotteixeira: After UI meeting

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -0,0 +1,173 @@
...
+
+import { makeStyles } from "@material-ui/styles";
+import searchAreaStyle from "./SearchArea.module.css";
+
+import SearchBar from "./SearchBar";
+import ShowAdvancedOptionsButton from "./ShowAdvancedOptionsButton";
+import SliderValueTooltip from "./SliderValueTooltip";
+
+const INITIAL_JOB_TYPE = "";
+const INITIAL_JOB_DURATION = 1;
```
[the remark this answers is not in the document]
imnotteixeira: We are mixing things, these values are for rendering purposes, regarding `INITIAL_JOB_TYPE`, it can start as `undefined`, but `INITIAL_JOB_DURATION` must have a value, as the slider needs it to render. However, that filter logic can and should be done in the `submitForm` function, that will handle the API calling and fetching the search results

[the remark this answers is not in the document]
imnotteixeira: I've put a comment in the function to remind the future implementer of this functionality about the filter nature of the fields, which are optional

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -0,0 +1,173 @@
...
+
+SearchArea.propTypes = {
+    addSnackbar: PropTypes.func,
+    onSubmit: PropTypes.func.isRequired,
+};
+
+export const mapStateToProps = () => ({
+});
+
+export const mapActionsToProps = (dispatch) => ({
```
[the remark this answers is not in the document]
imnotteixeira: <3

src/components/HomePage/SearchArea/jobTypes.js
```diff
@@ -0,0 +1,25 @@
...
+        value: "PART_TIME",
+        label: "Part-Time",
+    },
+    {
+        value: "FULL_TIME",
+        label: "Full-Time",
+    },
+
+]);
+
```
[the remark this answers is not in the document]
imnotteixeira: I don't think we should have an `OTHER` option as a filter, as it provides no information. Instead, we should only have concrete filters, and update them as we see fit/necessary over time.
Changing the value of ANY to `undefined`

package.json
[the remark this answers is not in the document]
miguelpduarte: Still, can these be safely removed? Maybe it would make sense to do so

src/App.js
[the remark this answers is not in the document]
miguelpduarte: > The idea was to have a Navbar in the future, but as the visual design planning phase was a bit sloppy, nothing is decided yet

Then just remove it instead of commenting it. If it is added in the future, then it will be added there. Otherwise, this will just be code clutter.

src/actions/notificationActions.js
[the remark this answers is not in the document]
miguelpduarte: Right, but stuff like `redux-actions` and similar make use of `payload`. If we eventually change to it, we would have to switch everything, and having stuff nested in the payload looks like a cleaner structure to me, you would always expect the `type` and `payload` keys, instead of varying ones.

src/components/HomePage/ShowMoreButton.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import { IconButton } from "@material-ui/core";
+import { KeyboardArrowDownRounded } from "@material-ui/icons";
+
+const ShowMoreButton = ({ className, onClick }) => (
```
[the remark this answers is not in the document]
imnotteixeira: I see, but what's your opinion on the styles location? Should I still keep them in the homePageStyles file? For me, I think it' s best because they are quite small and having a lot of small styling files clutters a bit, on the other hand, I like the separation of concerns :s

src/components/Notifications/Notifier.js
```diff
@@ -1,40 +1,44 @@
...
+        this.displayed = [...this.displayed, id];
     };
 
-    render() {
-        const { notifications, enqueueSnackbar, removeSnackbar } = this.props;
-        const { displayed } = this.state;
+    shouldComponentUpdate({ notifications: newSnacks = [] }) {
+        const { notifications: currentSnacks } = this.props;
+        for (let i = 0; i < newSnacks.length; i += 1) {
+            if (!currentSnacks.filter(({ key }) => newSnacks[i].key === key).length)
```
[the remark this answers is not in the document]
imnotteixeira: A plain ES6 `Set`? No. (https://github.com/reduxjs/redux/issues/1499) I tried it in another project and it was not a good idea. An alternative mentioned in the Redux docs is `Immutable.js` (https://redux.js.org/recipes/using-immutablejs-with-redux), but it presents a lot of pitfalls and requires careful programming because its structures are handled differently than normal JS Objects, so you need to know what you're doing basically. And, even if I know that me, or you, or 1 or 2 more contributors would be careful as well, and do their homework of reading the pitfalls and the How-To's, the rest won't, which will cause a lot of problems and anti-patterns, leading to inefficient solutions, inadvertently, and, as they say in the link, Immutable.js is hard to debug, because they are not normal Objects. So, I think I prefer this O(n^2).

src/index.js
```diff
@@ -1,14 +1,19 @@
+/* istanbul ignore file */
 import React from "react";
 import ReactDOM from "react-dom";
 import { BrowserRouter } from "react-router-dom";
 import { SnackbarProvider } from "notistack";
+import { ThemeProvider } from "@material-ui/styles";
+import AppTheme from "./AppTheme.js";
 
 import "./index.css";
```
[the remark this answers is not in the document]
imnotteixeira: This is setting the App Theme, which is different from the `CssBaseline` goal. Basically here we set the font family, sizes, the palette, etc... which is specific to this application. `CssBaseline` on the other hand, makes sure that the styling looks similar in all browsers, normalizing their presets. 

`CssBaseline` will be added

package.json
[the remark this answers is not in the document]
imnotteixeira: I think it can

src/pages/HomePage.js
```diff
@@ -1,21 +1,58 @@
...
+    });
+};
+
+const HomePage = () => {
+
+    const [productDescriptionRef, setProductDescriptionRef] = React.useState(null);
+    const [searchResultsRef, setSearchResultsRef] = React.useState(null);
+
+    // Setting the offers, probably needs re-thinking
+    // Maybe use React.useEffect for offer loading
```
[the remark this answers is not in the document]
imnotteixeira: I did some refactors regarding the `Offers`. The offers will be fetched once the user searches (might change later, if we want to show some preferred ones from the beginning), but the point is: they will be stored in the redux state, and this component won't be the one responsible for fetching them, the `SearchArea` will do it `onSubmit`. This component will connect to the store and pass the current offers to the `SearchResults`. If the offers array is empty (initial state), the `SearchResults` renders the skeleton version, to make it seem they are loading. However, this is not enough, because it does not make sense to have x offers loading without having made a search yet. So, the `SearchResults` has now a `show` flag, which is enabled when the user searches for something, triggering the render of the `SearchResults`component and, while the offers are not ready yet, there will be a skeleton, when they arrive, they will be shown.

src/pages/HomePage.js
```diff
@@ -1,21 +1,58 @@
...
+import SearchResults from "../components/HomePage/SearchResultsArea/SearchResults";
+import ProductDescription from "../components/HomePage/ProductPlacementArea/ProductDescription";
+
+export const scrollToProductDescription = (productDescriptionRef) => {
+    productDescriptionRef.current.scrollIntoView({
+        behavior: "smooth",
+        block: "start",
+    });
+};
+export const scrollToSearchResults = (searchResultsRef) => {
```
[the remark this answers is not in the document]
imnotteixeira: Moved to `utils/`

src/components/HomePage/SearchResultsArea/Offer/OfferCard.js
```diff
@@ -0,0 +1,86 @@
...
+
+const logo = require("./ni.png");
+
+const useStyles = makeStyles(OfferCardTheme);
+
+// THIS NEEDS RE-THINKING - POSSIBLE RE-DESIGN PENDING
+
+const OfferCard = (props) => {
+
+    // REFACTOR THIS INTO an Offer Object which could be easily extended and more testable
```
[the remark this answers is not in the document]
imnotteixeira: Refactor explained above

src/components/HomePage/SearchResultsArea/Offer/OfferCardTheme.js
```diff
@@ -0,0 +1,40 @@
...
+    },
+    details: {
+        display: "flex",
+        flexDirection: "column",
+        width: "100%",
+    },
+    logo: {
+        height: 100,
+        width: 100,
+        marginLeft: AppTheme.spacing(3),
```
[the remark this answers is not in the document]
imnotteixeira: Solved.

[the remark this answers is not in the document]
imnotteixeira: There was only one more case, `SearchBarTheme`

src/components/HomePage/ShowMoreButton.js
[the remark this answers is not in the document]
imnotteixeira: On further consideration, I think this makes sense, sort of, to be done like this. If you check the code, what the class is doing, is styling the wrapper div of the button, so that it can be used in more scenarios (you can customize the position, alignment, etc) but not to change the internal look of the button (I mean, it can, if you really want to, but that would be the same as having it in the general .css file and targeting that MUIButton class specifically. Therefore, I think we should keep it this way.

src/components/HomePage/MainView.js
```diff
@@ -0,0 +1,38 @@
...
+                    alt="nijobs Logo"
+                />
+            </div>
+        </div>
+        <InfoBox
+            info="Your next oportunity is out there. Use the search bar to find it!"
+        />
+        <SearchArea onSubmit={scrollToSearchResults}/>
+        <ShowMoreButton
+            className={homePageStyles.showMoreBtn}
```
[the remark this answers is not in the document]
imnotteixeira: Answered below.

.eslintrc
```diff
@@ -8,6 +8,9 @@
 
     "globals": {
         "jsdom": true,
+        "shallow": true,
+        "render": true,
+        "mount": true
```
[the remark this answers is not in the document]
imnotteixeira: From my research, it must be done like that, but if you find something it can always be changed later, not that it makes much of a difference, as this only affects the dev stage and doesn't impact the application

package.json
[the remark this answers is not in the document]
miguelpduarte: :shipit:

src/pages/HomePage.js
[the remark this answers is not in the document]
miguelpduarte: Makes sense!

src/components/HomePage/SearchArea/jobTypes.js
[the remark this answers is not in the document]
miguelpduarte: However, this disables the ability to search for something that is not specified in the current filter (it would only be searchable using `ANY`, which would show everything else). I don't think that this is very good but using `Other` would also be less than ideal, as it would be manual database migrating if other possibilities are added 🙃

src/components/HomePage/ShowMoreButton.js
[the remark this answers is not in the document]
miguelpduarte: I have no problem with several different (but related) components including the same file. I have one, however, with components passing styling concerns to children - imo each component should be concerned with its own styling, at most being able to pass some options (like bootstrap-style colors: `primary`, `secondary`, etc) but other than that not really messing with the look of their children.

src/components/Notifications/Notifier.js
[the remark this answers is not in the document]
miguelpduarte: Fair point, did not know about that. Thought that the ES6 `Set` could be used as a drop-in replacement.

As such, disregard the initial comment.

src/index.js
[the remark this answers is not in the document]
miguelpduarte: I was referring to index.css, is that also what you meant?

src/components/HomePage/SearchArea/jobTypes.js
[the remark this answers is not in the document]
imnotteixeira: I think the idea of having all the possibilities disclosed is the best. It removes ambiguity and allows for better searching. Regarding searching for something that is not specified in the current filter, it would not happen because everything is part of those options and, as I mentioned before, I can always add more options, but I don't think there are many more, so it would be a good trade-off

src/components/HomePage/ShowMoreButton.js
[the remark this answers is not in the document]
miguelpduarte: (The comment section didn't update with the new comment 🙃 )

Wouldn't it make more sense for the parent component to be responsible for wrapping the button and aligning it as necessary then?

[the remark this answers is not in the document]
imnotteixeira: That's what I was thinking now, I probably makes sense to do it that way instead

src/components/HomePage/SearchArea/jobTypes.js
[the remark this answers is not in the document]
miguelpduarte: Maybe not in the case for job types, but maybe in other things, I guess

src/components/HomePage/ShowMoreButton.js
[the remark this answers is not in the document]
miguelpduarte: :shipit:

src/index.js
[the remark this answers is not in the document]
imnotteixeira: No, I was referring to the AppTheme (as it was what Github highlighted from the commit addition), yeah I don't know if we need that index.css after adding the `CssBaseline`... Removed

src/components/HomePage/SearchArea/jobTypes.js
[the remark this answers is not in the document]
imnotteixeira: Yeah, there may be filters that need an `OTHER` value, but I think jobTypes is not one of them

src/index.js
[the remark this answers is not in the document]
miguelpduarte: Thanks GitHub! 🤦‍♂ 

Well, anyway, that's that! 👍

src/actions/notificationActions.js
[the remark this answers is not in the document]
imnotteixeira: For me, it just seems like overcomplicating what doesn't need to be overcomplicated. Seems we would be optimizing just for a package we probably won't use, and we should keep actions as they are: Objects with at least a `type ` property, being the additional properties defined by the dev depending on the needs of the action.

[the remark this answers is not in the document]
miguelpduarte: Sure, I can roll with that - it doesn't seem like it will make any big difference in the end, like you mentioned

PR #19, opened 2019-08-06
Remove dead comments from index

xRuiAlves:
I helped ❤️  🍝


miguelpduarte:
LGTM, but seems kinda irrelevant.
Deletes some docs, but nothing that cannot be googled anyway, I guess.

@imnotteixeira any opinions on this? Don't see why to merge it, and also don't see why not 🙃

miguelpduarte:
Did not know that was the case.

As such, will close this then.

PR #22, opened 2019-08-06
Improve eslint configuration

### LINKED ISSUE #16: Improve eslint rules
As per:

- https://github.com/NIAEFEUP/nijobs-fe/pull/13#discussion_r310809717
- https://github.com/NIAEFEUP/nijobs-fe/pull/13#discussion_r310813247
- And probably some more

To do:

- [x] Add [linting for hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks)
- [x] Improve spacing rules
- [ ] Find a reliable way of managing components props (https://github.com/NIAEFEUP/nijobs-fe/pull/13#discussion_r310813247)

miguelpduarte:
Closes #16


miguelpduarte:
IMO this is ready to go

Would only require two opinions:

- Should we take the chance and add something like [Stylelint](https://stylelint.io/)? (Looks fancy)
- In my opinion, it might be better to split the testing code from the normal code, so I would propose moving `test` one directory up (so it is on the same level as `src`). This would make the docker images smaller as tests grow, and it is unnecessary clutter in production. Asking about this now because the `lint` command should also lint the tests and if they are to be moved, this should be taken into consideration (did not add it there yet because the command fails on inexistent directories)

miguelpduarte:
Also, should probably copy all but the react specific ones to the backend as well 😄 
Double the effect, with (nearly) no additional effort, yay!

xRuiAlves:
LGTM @miguelpduarte , but no permissions to approve :sweat_smile:

miguelpduarte:
@xRuiAlves you are in the org. Thus, you should have permissions.

Still, I'd rather wait on @imnotteixeira, since he's owning most of the code for now

xRuiAlves:
:man_facepalming: I was being dumbo. Nevertheless, I'm approving it. We can still wait on @imnotteixeira approval before merging

miguelpduarte:
> Can you 'uncommit' the src files that we're changed due to the new rules? I think this will collide with some of the ongoing PRs, and as it is automatic, we could just force commit the new eslint rules (avoid the linting) and when the merges occur for the other changes, they will be fixed with the --fix command.

Sure, I understand the concern for unnecessary conflicts! The git hooks configured via `husky` made me commit those, really. But I'll just ammend the commit and force push with the ignore hooks option or whatever it was (I'll find it 😅).


> Regarding tests, I think we should keep them aside the tested components, as in: if we have a component1.js, we will have a component1.spec.js or component1.test.js. This is because we will have mainly unit tests, which don't impact many components at the same time, and make a pair with the component they test. Also, with the separation of components that was started in #13, each component folder may have only a few files along with their test sibling files.

Hmm... I'm okay with doing this for now, but am not sure if this is the best way going forward. I understand the idea but am not totally sold on this structure (as there will always be some unit tests not related to the components, and those would be in a `test` directory somewhere, which would break this logic and scatter the test code).
I'd rather go with keeping the separate `test` directory and just mirroring the structure there, having a `components` folder, `lib` folder (if one starts existing), et cetera. This would keep the structure similar, and the names of the test files would mirror the actual code files. Imo, this is the best of both worlds.

> Also, doing it like this would avoid the giant import commands that traverse a lot of folders, as they would be well apart from the components.

I don't think that this is a problem, really. We can also look into just using something like [absolute imports](https://hackernoon.com/absolute-imports-with-create-react-app-4c6cfb66c35d) - Our version of `react-scripts` is 3+ iirc so it should work. Will investigate some linting rules for this.
It will seem weird at first, but I think this is a good idea going forward, the code will be much clearer on its desired behaviour. What do you think?

miguelpduarte:
@imnotteixeira Extraneous files removed. Can you please give some input to what I mentioned, especially about the absolute import paths and using stylelint

package.json
```diff
@@ -20,7 +20,7 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint \"**/*.js\" --fix"
+    "lint": "eslint src --fix"
```
imnotteixeira: How would this work in case the tests are moved away from the `src` components? Being actual functions and code, they should be linted as well as they also need to be maintained.

miguelpduarte: Just change it to `eslint src test --fix`

miguelpduarte: I didn't change the command yet because it would fail on an empty directory.

Check the commit message of 00d0a7adf7ab9ff28e6a401bec44bf3b3296b209

imnotteixeira:
```suggestion
    "lint": "eslint src test --fix"
```
Do code suggestions still work?

miguelpduarte: They do, but this will break that command.

(Also, something I just though of - shouldn't we have a `lint` command that didn't `--fix` for CI?)

imnotteixeira: There is a `test` folder inside the `src` folder, you can move that folder one level above.

Also regarding CI, I thought that already existed, but it seems not, you can add it

miguelpduarte: I know I could, but I didn't know if I should, hence my previous questions 😅 
Going to do that now, then.

Yup, sounds good. Will add it and rename this one to `lint-fix` most likely

imnotteixeira: Don't forget to make the -fix one the default when pre-committing to make our lives easier!

miguelpduarte:
> Regarding stylelint, I don't think it is really necessary, but you can add it if you want to play with it, I won't oppose.

I've got some pretty good configs so I think I'll give it a whirl.

> Regarding absolute imports, I would like them, assuming they work with our config

From what I've read they should be available in our version of `react-scripts`. But will try them and report back.

package.json
miguelpduarte: > Also regarding CI, I thought that already existed, but it seems not, you can add it

I think that it does not exist since there is no CI configured atm, see #25

miguelpduarte:
After all, configuring `stylelint` seems like it will be annoying and will not have much gain (I hope, if so this is a good sign).
Will instead focus on the aboslute imports part so we can close this PR, and then I'll work on adding CI (probably via CircleCI, since it's being set up in `niployments` currently as well and it's the CI tool I've seen the most recently).

miguelpduarte:
After investigating some more found the [react-scripts docs on testing](https://facebook.github.io/create-react-app/docs/running-tests#filename-conventions) which specify that they only scan for test files inside `src`.
Thus, I think that the absolute imports idea is now pointless and would like to focus on merging this ASAP so that I can work on #25

miguelpduarte:
Sorry for the force push spam, derped while squashing commits 😅 

I think this is ready to be 🚢

miguelpduarte:
Agreed! Yay for knowledge and discussion!

Gonna :shipit: then!

Issue #21, opened 2019-08-06
Update to node 10.16.0


miguelpduarte:
Should move to the LTS version.

Change Travis configs, docker stuff and add a `engine` specification in `package.json` to match the version ([see this](https://docs.npmjs.com/files/package.json#engines)).

Also, can we switch to using an `alpine` docker image? I think so, and this would probably be better due to it being smaller iirc

PR #24, opened 2019-08-06
Upgrade to node 10.16.0

xRuiAlves:
Closes https://github.com/NIAEFEUP/nijobs-fe/issues/21

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:10.16.0-alpine
 
 RUN mkdir -p /usr/src/nijobs-fe
 WORKDIR /usr/src/nijobs-fe
--- a/package-lock.json
+++ b/package-lock.json
@@ -4399,15 +4399,70 @@
       }
     },
     "eslint-plugin-react": {
-      "version": "7.11.1",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-react/-/eslint-plugin-react-7.11.1.tgz",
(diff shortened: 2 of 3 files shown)


miguelpduarte:
Please change the PR description to `closes` or something similar so GitHub picks it up.

LGTM, but don't feel okay approving this, paging @imnotteixeira 

The deploy seems to be fine, and I believe that the version was changed everywhere.
Do we not have a Travis or something to run the tests and ensure nothing broke? Thought we did, strange.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:10.16.0-alpine
```
imnotteixeira: Did you test this Docker config? I've had some problems with alpine images in the past due to some packages requiring a more complete version of node.

xRuiAlves: @imnotteixeira it looked fine, nothing un-usual detected. But if you could double check it, it would probabbly be best.

miguelpduarte: Just for reference: The node version is exactly the same - only the underlying OS is much smaller. See [this here](https://github.com/alpinelinux/docker-alpine#why) for more information.

imnotteixeira: I know, the problem was not with the node itself, but with some packages or dependencies that were required for some package installation

miguelpduarte: Was just making it clear for anyone that might read the discussion 😅 

I'm guessing it would be something like `sass` or similar, I think some of these engines use `make` or `cmake` or stuff like that so they would most likely break, yes.

miguelpduarte:
Can we merge this too? @imnotteixeira

imnotteixeira:
🚀Go Go Go 🚀

PR #26, opened 2019-08-09
Add CircleCI configuration

### LINKED ISSUE #25: Configure CI

miguelpduarte:
Closes #25

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,79 @@
+version: 2
+# Aliases to keep the configuration DRY (Don't Repeat Yourself)
+# '&' captures an alias and '*' expands it
+# '<<:' is used when you need to expand a level of block mapping, basically
+aliases:
+    - &docker_image
+        docker:
+            - image: node:10.16.0-alpine
+    - run: &setup_certificates 
+            # This is necessary to be able to restore from and save to CircleCI's cache and workspace
+            name: Install CA certificates
+            command: apk add ca-certificates
+    - &attach_workspace
+        attach_workspace: 
+            at: .
+jobs:
+    build:
+        <<: *docker_image
+        steps:
+            - run: *setup_certificates
+            - checkout
+            - restore_cache:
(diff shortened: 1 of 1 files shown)


miguelpduarte:
Force pushed because the config didn't apply (the check was not running).

miguelpduarte:
Well... Working as expected :sweat_smile:

@imnotteixeira how to act on this? You requested that I didn't update the code due to superfluous merge conflicts. However, I am now unsure on how to proceed now.

Should we make the CI check not necessary to merge (or just admin force-merge stuff) and merge this, making everything depend on #10 instead (to where `develop` will be rebased and the files finally linted)?
At the moment it seems like the best option to me, despite it feeling kind of weird to block everything else on a certain PR - not like there's much more activity going on here anyway :sweat_smile:

miguelpduarte:
Also, probably adding a badge would be cool too, [link here to the circleci one in the already used things](https://shields.io/category/build).

miguelpduarte:
Also, not sure if there is any point in parallelizing the jobs in a workflow, since we only get one container for execution anyway (no jobs will run in parallel).
I think it kind of just adds clutter to the PR checks :confused:

miguelpduarte:
> 
> 
> Also, not sure if there is any point in parallelizing the jobs in a workflow, since we only get one container for execution anyway (no jobs will run in parallel).
> I think it kind of just adds clutter to the PR checks, sadly 😕

Now I think I'm mistaken, as the logs of all of the jobs after `build` have the same timestamp for step `Spin up Environment`.
Time will tell, but I now I believe the jobs are better split 😅 
If not for anything else, this should make for cleaner config code I guess (and added some aliases at the top of the file now for the same reason).

miguelpduarte:
I was thinking of forcing that to reduce the friction of rebasing these changes into #13, but can fix the problems too, I think something like `npm audit fix` will suffice.

miguelpduarte:
🚢 'ing it

PR #33, opened 2019-11-07
Add Dockerfile-prod for deployment via niployments

### LINKED ISSUE #32: Add Dockerfile-prod
In order for this app to be automatically deployed via [niployments](https://github.com/NIAEFEUP/niployments), there needs to be a Dockerfile-prod file that builds the application and serves it.

imnotteixeira:
Closes #32

Added Dockerfile based on the template available on niployments for react applications, and added specific folders to the image (config/ and scripts/). 

.env not added because it's not being used for now and is not necessary at the moment. In the future, this might be updated.

The current image serves the build with nginx on port 80 (default) - this needs to be remapped when running probably

### DIFF
--- a/Dockerfile-prod
+++ b/Dockerfile-prod
@@ -1,0 +1,29 @@
+#Based on the template from niployments (https://github.com/NIAEFEUP/niployments) for react applications
+
+# First, building the static files
+FROM node:10.17.0-alpine as build
+
+WORKDIR /usr/src/app
+COPY package.json package-lock.json ./
+
+# Because colors break logs
+ENV NPM_CONFIG_COLOR=false
+
+# Production or not doesn't really matter as this image will not be used other than for building
+RUN npm install
+
+# Necessary files for building the app
+COPY public/ public/
+COPY config/ config/
+COPY scripts/ scripts/
+COPY src/ src/
+
+# Building the image
+RUN npm run build
+
+# Then, use nginx to serve the built files
+# See https://hub.docker.com/_/nginx
+FROM nginx:alpine
+
+# Copying the built files into the nginx image, to the default served directory
+COPY --from=build /usr/src/app/build /usr/share/nginx/html
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "engines": {
-    "node": "10.16.0"
+    "node": "10.17.0"
   },
   "dependencies": {
     "@babel/core": "7.5.5",


miguelpduarte:
> The current image serves the build with nginx on port 80 (default) - this needs to be remapped when running probably

Yup, this is how niployments is currently working: Since custom nginx configs are necessary for port remapping, it is preferrable to just have the docker containers exposing port 80, and that is remapped when running with `docker run -p 80:<desired port>` (order might be wrong, do not have the docs at hand).

Dockerfile-prod
```diff
@@ -0,0 +1,34 @@
+#Based on the template from niployments (https://github.com/NIAEFEUP/niployments) for react applications
+
+# First, building the static files
+FROM node:10.16.2-alpine as build
```
miguelpduarte: `10.16.3`? Also, it seems that the latest LTS is already `12.13.0`. However that might have breaking changes, we need to analyze it. Still, the latest version for the dubnium LTS (10.x) is `10.17.0`.

I believe that we should bump this version everywhere to at least latest dubnium and later on probably latest LTS, erbium (`12.13.0`).

Dockerfile-prod
```diff
@@ -0,0 +1,34 @@
...
+
+# Because colors break logs
+ENV NPM_CONFIG_COLOR=false
+
+# Production or not doesn't really matter as this image will not be used other than for building
+RUN npm install
+
+# Necessary files for building the app
+COPY public/ public/
+
```
miguelpduarte: Nitpick: Remove newline?

Dockerfile-prod
```diff
@@ -0,0 +1,34 @@
...
+# Building the image
+RUN npm run build
+
+# Then, use nginx to serve the built files
+# See https://hub.docker.com/_/nginx
+FROM nginx:alpine
+
+# Copying the built files into the nginx image, to the default served directory
+COPY --from=build /usr/src/app/build /usr/share/nginx/html
+# If a custom nginx.conf is necessary (for routing some stuff, sometimes React Router has trouble with nginx it seems - should use @reach anyway tbh)
```
miguelpduarte: Maybe the comments downwards from here can be removed as they do not add anything. They were present in the niployments Dockerfile-react template mostly for documentation, but they are not necessary here imo.

miguelpduarte:
Renamed, triggered

Dockerfile-prod
imnotteixeira: We could just rename to node:10-alpine just to make sure it stays up to date within v10

miguelpduarte: `dubnium-alpine`? Idk, it feels strange to trust in automatic bumping. I know it will be inside the same major release, but still...

imnotteixeira: Sure, I will edit the `engines` field in package.json as well

miguelpduarte: 👍

miguelpduarte:
Merging this should trigger a deployment that will (hopefully) work!

PR #38, opened 2019-11-10
Add nginx config and update dockerfile to consume it

imnotteixeira:
Currently, when manually going to `mainURL`/something will give a 404 nginx error if the page does not exist, however, with this PR, it will redirect to index.html, so that react-router can deal with the route inside it

### DIFF
--- a/Dockerfile-prod
+++ b/Dockerfile-prod
@@ -24,12 +24,10 @@ COPY .env* ./
 # Building the image
 RUN npm run build
 
-# Copy env files
-COPY .env* ./
-
 # Then, use nginx to serve the built files
 # See https://hub.docker.com/_/nginx
 FROM nginx:alpine
 
 # Copying the built files into the nginx image, to the default served directory
 COPY --from=build /usr/src/app/build /usr/share/nginx/html
+COPY nginx.nijobs.conf /etc/nginx/conf.d/default.conf
--- a/nginx.nijobs.conf
+++ b/nginx.nijobs.conf
@@ -1,0 +1,36 @@
+server {
+        # listen on port 80
+        listen 80;
+        # save logs here
+        access_log /var/log/nginx/access.log main;
+        # where the site files are
+        root /usr/share/nginx/html;
+        # what file to serve as index
+        index index.html index.htm;
+
+        location / {
+            # First attempt to serve request as file, then
+            # as directory, then fall back to redirecting to index.html
+            try_files $uri $uri/ /index.html;
+        }
+
+        # Media: images, icons, video, audio, HTC
+        location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc)$ {
+          expires 1M;
+          access_log off;
+          add_header Cache-Control "public";
+        }
+
+        # Javascript and CSS files
+        location ~* \.(?:css|js)$ {
+            try_files $uri =404;
+            expires 1y;
+            access_log off;
+            add_header Cache-Control "public";
+        }
+
+        # Any route containing a file extension (e.g. /devicesfile.js)
+        location ~ ^.+\..+$ {
+            try_files $uri =404;
+        }
+    }


Dockerfile-prod
```diff
@@ -24,12 +24,10 @@ COPY .env* ./
 # Building the image
 RUN npm run build
 
-# Copy env files
```
miguelpduarte: Why was this removed?

miguelpduarte: Hmm, this seems to be the old wrongly placed one, but was this not already removed?

Dockerfile-prod
```diff
@@ -24,12 +24,10 @@ COPY .env* ./
...
-COPY .env* ./
-
 # Then, use nginx to serve the built files
 # See https://hub.docker.com/_/nginx
 FROM nginx:alpine
 
 # Copying the built files into the nginx image, to the default served directory
 COPY --from=build /usr/src/app/build /usr/share/nginx/html
+COPY ./nginx-config/nijobs.conf /etc/nginx/conf.d/default.conf
```
miguelpduarte: Why have a directory? Also, the `./` in the copy from looks weird, and is unnecessary.
The reason would be to override all files in the directory, but that only applies using volumes so it would be the same to `RUN rm /etc/nginx/conf.d/default.conf` and then `COPY nijobs-nginx.conf /etc/nginx/conf.d/whatever.conf`

Dockerfile-prod
imnotteixeira: I did a git pull on develop before this, so it must have slipped, it's no big deal actually

Dockerfile-prod
miguelpduarte: Why have a directory? Also, the `./` in the copy from looks weird, and is unnecessary.
The reason I see would be to override all files in the directory, but that only applies using volumes so it would be the same to just `RUN rm /etc/nginx/conf.d/default.conf` and then `COPY nijobs-nginx.conf /etc/nginx/conf.d/whatever.conf`

nginx-config/nijobs.conf
```diff
@@ -0,0 +1,37 @@
+server {
+        # listen on port 80
+        listen 80;
+        # save logs here
+        access_log /var/log/nginx/access.log main;
+
+        # where the root here
+        root /usr/share/nginx/html;
```
miguelpduarte: This can be excluded, as it is the default, I believe

nginx-config/nijobs.conf
```diff
@@ -0,0 +1,37 @@
...
+        location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc)$ {
+          expires 1M;
+          access_log off;
+          add_header Cache-Control "public";
+        }
+
+        # Javascript and CSS files
+        location ~* \.(?:css|js)$ {
+            try_files $uri =404;
+            expires 1y;
```
miguelpduarte: Is this desired? Will new versions of files hit cache and thus not update?

nginx-config/nijobs.conf
```diff
@@ -0,0 +1,37 @@
...
+            # as directory, then fall back to redirecting to index.html
+            try_files $uri $uri/ /index.html;
+        }
+
+        # Media: images, icons, video, audio, HTC
+        location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc)$ {
+          expires 1M;
+          access_log off;
+          add_header Cache-Control "public";
+        }
```
miguelpduarte: Is this necessary? Can't we let nginx/the browser decide defaults?

nginx-config/nijobs.conf
```diff
@@ -0,0 +1,37 @@
+server {
+        # listen on port 80
+        listen 80;
+        # save logs here
+        access_log /var/log/nginx/access.log main;
+
+        # where the root here
+        root /usr/share/nginx/html;
+        # what file to server as index
```
miguelpduarte: typo

nginx-config/nijobs.conf
```diff
@@ -0,0 +1,37 @@
+server {
+        # listen on port 80
+        listen 80;
+        # save logs here
+        access_log /var/log/nginx/access.log main;
+
+        # where the root here
```
miguelpduarte: typo?

nginx-config/nijobs.conf
imnotteixeira: From what I've seen, a lot of examples had /var/www as default, we should keep it in order to make it clearer

miguelpduarte: The fact that the Dockerfile worked without an nginx.conf disproves that. The default for nginx's static file serving is `/usr/share/nginx/html`

nginx-config/nijobs.conf
imnotteixeira: I might be wrong, but in this setup, if there are "new files" there will be another container generated, therefore another nginx instance and no cache problems

miguelpduarte: This cache is in the browsers. This is the header setting for that expiration, if I'm not mistaken

miguelpduarte: It would make no sense for nginx to serve outdated versions of local files 🙃 the overhead would be bigger on creating that cache system than just serving them outright

Dockerfile-prod
imnotteixeira: It is not necessary to have a rm + cp. Instead we just cp to the default, which will override and work, since it's in an isolated environment, it won't cause a problem when overriding.

nginx-config/nijobs.conf
imnotteixeira: That might be true, however, it doesn't hurt to have it there to make it clearer for future devs.

nginx-config/nijobs.conf
imnotteixeira: I've been researching, and this is actually useful. Every time we build and there is a change in the files, they change their name (they are appended with some kind of hash I believe). Therefore, if a build only changes some files that might only update some built ones and keep the others in cache, which is good

Dockerfile-prod
miguelpduarte: Sure. However, using a directory just for holding an nginx conf file seems weird to me

nginx-config/nijobs.conf
miguelpduarte: Sure, I'm fine with having it this way

nginx-config/nijobs.conf
imnotteixeira: This ensures consistency and a good experience for everyone. I don't know the default expiry date for images, but considering that the only images we'll have are either static (our logo, some helper svgs, ...) or almost static (companies profile logos - and even these will have an appended id, making sure that if they are updated, they change) I think that 1Month is a good value.

Regarding the Cache-Control: public - it's actually better since it will allow us to cache these images even with https, if it were private, which is the default, the images wouldn't be cached

This is a really good doc on web caches: https://www.mnot.net/cache_docs/

nginx-config/nijobs.conf
miguelpduarte: Okay, good to go then!

nginx-config/nijobs.conf
miguelpduarte: Sounds good then!

PR #39, opened 2019-11-13
Add variable basename on router based on env var

imnotteixeira:
This allows us to set up the base url of different environments without breaking the router (prod will be on ni.fe.up.pt/nijobs, but dev will be on ni.fe.u.pt/nijobs/st4g1ng)


miguelpduarte:
> 
> 
> This allows us to set up the base url of different environments without breaking the router (prod will be on ni.fe.up.pt/nijobs, but dev will be on ni.fe.u.pt/nijobs/st4g1ng)

dev -> ni.fe.up.pt/st4g1ng/nijobs actually, but yes

PR #41, opened 2019-11-18
Added a flex wrapper to center SearchArea in MainView

imnotteixeira:
This is something that was not easily noticeable, unless you use uncommon resolutions (my viewport is usually 1920*2160 - like two 1080p screens on top of each other), which made the search area go to places where it shouldn't.

It should be fixed now, I simply added a flex wrapper that centers the search area vertically.
---
Before:
nijobs_b4

After:
nijobs_after


imnotteixeira:
Done in #61

PR #42, opened 2019-11-18
Add Suspense to Router and Lazy loading of page components

imnotteixeira:
This change takes advantage of the recent React.lazy and React.Suspense APIs which allow us to load our components more selectively, relieving the end user from loading components he is not even looking at.

Extra: Added progress bars between page changes, it looks so good.


imnotteixeira:
I would like to highlight that the Suspense API (required to use `React.Lazy()`) is still experimental, however, the usage of the lazy feature alone should be ok, the concurrent mode is the one that is still not ready yet. So this should be taken into account before merging.

Issue #7, opened 2019-01-25
Register and Login pages


miguelpduarte:
Pages to add:
* Register
* Login
* Apply for our site

The login page should have the following inputs:
* Email
* Password

The register page should have the following inputs (should be able to receive the one-time token from the URL - not yet in use):
* Email
* Nome da empresa
* Password

The apply for our site page should have the following inputs:
* Email
* Nome da empresa
* Motivação
* Outros contactos

imnotteixeira:
After further discussion, this is no longer the required work. Here is the updated version of this issue:

Pages to add:

- Complete registration (the approved users already have an account, so they will need to provide a company logo and bio on first login - they can also update their password if desired )
- Login

The login page should have the following inputs:
* Email
* Password

The register part should be a progress-like step-by-step process of completing the missing information. Steppers are a nice touch (https://material-ui.com/components/steppers/)

imnotteixeira:
Also, as discussed offline with @miguelpduarte, we will implement the login state details using SWR (https://swr.now.sh/), as it can be a much-requested piece of information by various parts of the application, thus using a cache-ready approach seems reasonable.

PR #49, opened 2020-03-11
Add User session functionality

imnotteixeira:
Contributes to #7 

This PR adds the possibility of:
* login
* ~~register~~ (This is done as part of the Company Applications here: #48)
* get current session info

It uses an SWR hook to fetch the user session, while caching the result so that the API doesn't get flooded with requests everytime a page/component requires session details

**EDIT**: It actually, makes the request every time, but uses the current data (from previous request) while the up to date hasn't come. That's where the _Stale While Revalidate_ comes from...


imnotteixeira:
### Desktop UX (1080p)

user-menu-desktop-ux

### Mobile UX (iPhone 5/SE - One of the smallest among the common ones)

user-menu-mobile-ux

PR #66, opened 2021-01-24
Add devtools UI to allow API host changing

imnotteixeira:
Adds a "DevTools" section to allow changing the API host, so that we can test functionalities for example in netlify with custom backends not yet developed.

Only applies in all envs but `production`, unless `ALLOW_DEV_TOOLS` env variable is set to true

image


src/index.js
```diff
@@ -14,21 +14,25 @@ import App from "./App";
...
+import { loadDevTools } from "./devTools.js";
 
 const jss = create({
     plugins: [...jssPreset().plugins, compose()],
 });
 
-
-ReactDOM.render(
-    <BrowserRouter>
+loadDevTools().then(() => {
```
monkin77, line 23: Why is the DevTools bar loaded like this instead of including it inside the App Component?

src/services/applicationsReviewService.spec.js
```diff
@@ -1,4 +1,6 @@
-import { API_HOSTNAME } from "../config";
+import config from "../config";
+const { API_HOSTNAME } = config;
```
monkin77, line 2: does this variable change when the api host changes?

monkin77 (commented):
Some questions relative to this pr

monkin77 (approved):

src/AppTheme.js
```diff
@@ -24,11 +21,16 @@ const theme = responsiveFontSizes(createMuiTheme({
...
-export default theme;
 
-export const primaryColor = "#C03736";
-export const secondaryColor = "#4F1315";
-export const tertiaryColor = "#C1D2D6";
-export const darkTextColor = "#2C3031";
+// Add custom palette variants
+theme.palette = {
+    ...theme.palette,
+    tertiary: theme.palette.augmentColor({
```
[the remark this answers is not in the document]
imnotteixeira: It basically picks up the main color and generates the light, dark and contrastText colors based on it. It does that automatically for the primary and secondary colors we define abov, but not for the others, that's why I separated them here so I can call that manually

src/hooks/useSession.js
```diff
@@ -1,6 +1,8 @@
 import useSWR, { mutate } from "swr";
 
-import { API_HOSTNAME } from "../config";
+import config from "../config";
+const { API_HOSTNAME } = config;
+
```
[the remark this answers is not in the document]
imnotteixeira: It's a React hook that components can use to get the current user session. If a user is logged in, it will return data for that user, such as email, company details (if applicable), etc..

src/index.js
imnotteixeira: First. it's a matter of semantics. Dev tools are not a part of the application, so they shouldn't be there.

Second, it's one component less to handle when devtools are disabled 

Third, doing it this way, allows this logic (to activate devtools or not) to run on build time, instead of running on clients locally every time they render a page

src/services/applicationsReviewService.spec.js
imnotteixeira: The variable itself changes, however the imports are constant (even though the variable in `../config`has changed, the variable inside this file will not update. That's why a refresh is required on change, so that the devtools load the updated value from the localStorage and apply it

PR #68, opened 2021-02-05
Feature/contact

monkin77:
Added a contact section to the home page


imnotteixeira (changes requested):
Visually, this is done. I just have some comments regarding the code itself.

Also you need to write the tests, we use react-testing-library, and in this case asserting that there are links for each thing should be enough.
Also add a test to the homepage to assert that the contact section shows

src/components/HomePage/ContactSection.spec.js
```diff
@@ -0,0 +1,59 @@
...
+// import HomePage from "../../pages/HomePage";
+
+afterEach(cleanup);
+
+describe("ContactSection", () => {
+    describe("render", () => {
+        it("renders without crashing", () => {
+            const div = document.createElement("div");
+            ReactDOM.render(<ContactSection />, div);
+        });
```
imnotteixeira, line 14: You don't need this test, if it somehow fails, the other tests will fail as well

src/components/HomePage/ContactSection.spec.js
```diff
@@ -0,0 +1,59 @@
+import React from "react";
+import ContactSection from "./ContactSection";
+import ReactDOM from "react-dom";
+import { render, cleanup } from "@testing-library/react";
+// import HomePage from "../../pages/HomePage";
+
+afterEach(cleanup);
```
imnotteixeira, line 7: The rtl docs state the following:

> Please note that this is done automatically if the testing framework you're using supports the afterEach global and it is injected to your testing environment (like mocha, Jest, and Jasmine). If not, you will need to do manual cleanups after each test.

So you should remove this line to avoid confusion in the future

src/components/HomePage/ContactSection.spec.js
```diff
@@ -0,0 +1,59 @@
...
+            const { queryByTitle } = render(<ContactSection />);
+            const btn = queryByTitle("twitter");
+            expect(btn.getAttribute("href").length > 0);
+        });
+
+        it("check linkedin button link", () => {
+            const { queryByTitle } = render(<ContactSection />);
+            const btn = queryByTitle("linkedin");
+            expect(btn.getAttribute("href").length > 0);
+        });
```
imnotteixeira, line 57: Move these tests to the render describe and instead of checking the length of href, assert its values to be correct

imnotteixeira (changes requested):

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,148 @@
...
+            <Grid container spacing={0}  >
+                <Grid item xs={12} sm={6} md={3}>
+                    <List>
+                        <ListItem>
+                            <Typography variant="h5">Contact us</Typography>
+                        </ListItem>
+                        <ListItem>
+                            <Button
+                                href="mailto:ni@aefeup.pt"
+                                startIcon={<Email htmlColor={Theme.palette.primary.main} />}
```
imnotteixeira, line 56: Don't import the theme directly, there can be overrides in place which won't work this way. Instead, you can use the `useTheme()` hook which returns the theme in place here, and you should use that instead.

You can declare it before (similar to the useStyles call), and reuse it whenever you need it in the tree

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,148 @@
...
+                    <Grid container spacing={0} >
+                        <Grid item xs={6} md={4}>
+                            <List>
+                                <ListItem>
+                                    <Typography variant="h5">Social</Typography>
+                                </ListItem>
+                                <ListItem dense>
+                                    <Button
+                                        href="https://ni.fe.up.pt"
+                                        startIcon={<Language htmlColor={Theme.palette.tertiary.main} />}
```
imnotteixeira, line 75: same thing for the theme hook

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,148 @@
...
+                                        className={classes.colorText}
+                                        title="website"
+                                    >
+                                        <Typography variant="button">Website</Typography>
+                                    </Button>
+                                </ListItem>
+                                <ListItem>
+                                    <Button
+                                        href="https://facebook.com/NIAEFEUP/"
+                                        startIcon={<Facebook htmlColor={Theme.palette.tertiary.main}/>}
```
imnotteixeira, line 85: theme comment

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,148 @@
...
+                                        className={classes.colorText}
+                                        title="facebook"
+                                    >
+                                        <Typography variant="button">Facebook</Typography>
+                                    </Button>
+                                </ListItem>
+                                <ListItem dense>
+                                    <Button
+                                        href="https://github.com/NIAEFEUP"
+                                        startIcon={<GitHub htmlColor={Theme.palette.tertiary.main} />}
```
imnotteixeira, line 95: theme comment

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,148 @@
...
+                        </Grid>
+                        <Grid item xs={6} md={4}>
+                            <List>
+                                <ListItem>
+                                    <Typography variant="h5" className={classes.backgroundColor}>.</Typography>
+                                </ListItem>
+                                <ListItem dense>
+                                    <Button
+                                        href="https://www.instagram.com/niaefeup/"
+                                        startIcon={<Instagram htmlColor={Theme.palette.tertiary.main} />}
```
imnotteixeira, line 112: theme comment

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,148 @@
...
+                                        className={classes.colorText}
+                                        title="instagram"
+                                    >
+                                        <Typography variant="button">Instagram</Typography>
+                                    </Button>
+                                </ListItem>
+                                <ListItem>
+                                    <Button
+                                        href="https://twitter.com/niaefeup"
+                                        startIcon={<Twitter htmlColor={Theme.palette.tertiary.main}/>}
```
imnotteixeira, line 122: theme comment

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,148 @@
...
+                                        className={classes.colorText}
+                                        title="twitter"
+                                    >
+                                        <Typography variant="button">Twitter</Typography>
+                                    </Button>
+                                </ListItem>
+                                <ListItem dense>
+                                    <Button
+                                        href="https://pt.linkedin.com/company/nifeup"
+                                        startIcon={<LinkedIn htmlColor={Theme.palette.tertiary.main} />}
```
imnotteixeira, line 132: theme comment

src/components/HomePage/ContactSection.spec.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import ContactSection from "./ContactSection";
+import { render } from "@testing-library/react";
+
+describe("ContactSection", () => {
+    describe("render", () => {
+        it("check mail button link", () => {
+            const { queryByTitle } = render(<ContactSection />);
+            const btn = queryByTitle("email");
+            expect(btn.getAttribute("href").match("mailto:ni@aefeup.pt"));
```
imnotteixeira, line 10: You are using regex to match the string, but you can use the string comparison directly, which is probably more performant.

Additionally, you are not "expecting" anything, since you need a matcher after the expect (https://jestjs.io/docs/en/expect#expectvalue)

The tests (and this goes for all the tests you have here -I won't repeat this comment) should follow something like this:

```js
expect(btn.getAttribute("href")).toBe("mailto:ni@aefeup.pt");
```

imnotteixeira (changes requested):

src/components/HomePage/ContactSection.js
```diff
@@ -0,0 +1,149 @@
...
+        padding: theme.spacing(4),
+    },
+    colorText: {
+        color: Theme.palette.tertiary.main,
+    },
+    backgroundColor: {
+        color: Theme.palette.dark.main,
+    },
+    primaryColor: {
+        color: Theme.palette.primary.main,
```
imnotteixeira, line 36: You're receiving the theme as argument to be able to use it inside, don't import the theme manually

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #69, opened 2021-02-08
Added GithubActions test workflow

imnotteixeira:
Adding github action to replace circle ci eventually


imnotteixeira:
Merging since the actions ran, which is the goal of this PR, for now

PR #70, opened 2021-02-08
Removed CircleCI config

imnotteixeira:
Since #69, we are using Github Actions. And since they allow for better running times, we are replacing CircleCI with them.

Also, currently, memory constraints in CircleCI are preventing our tests to run (and pass) as they do on local machines. Hence the need for change.


andrefmrocha:
Circle CI builds have been stopped. Try to push a ghost commit

PR #72, opened 2021-02-12
Feature/page layout

### LINKED ISSUE #71: Add Page Layout
Create a page layout that will render all pages with the footer.

monkin77:
Closes #71


src/components/PageLayout.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import ContactSection from "./HomePage/ContactSection";
+import propTypes from "prop-types";
+import Navbar from "./Navbar/index";
+
+const PageLayout = (props) => (
```
imnotteixeira, line 6:
```suggestion
const PageLayout = ({children}) => (
```

src/components/PageLayout.js
```diff
@@ -0,0 +1,19 @@
...
+    <div>
+        <Navbar showHomePageLink={props.showHomePageLink} />
+        {props.children}
```
imnotteixeira, line 9:
```suggestion
        {children}
```

imnotteixeira (changes requested):

imnotteixeira (commented):
image

(Highlighted text to better see what is going on since the letters were white-on-white)

The navbar is broken on mobile for pages that use full-page dialogs (it is showing on top of them)

I don't know what's the best approach here, but in these cases probably adding a single Home icon on the left of the title is a good start

imnotteixeira (changes requested):
The footer on non-Homepage pages is not quite right

image

Notice that the links are not aligned, for some reason

Also, on the homepage mobile version, the hamburger menu has a secondary color, and it should be white

imnotteixeira (changes requested):
Please unstage the package-lock.json since the audit problems remain, we will deal with them later.

After that, I am approving :)

imnotteixeira (approved):

PR #75, opened 2021-03-01
Added MyOffers page, the respective route and a Link on the navbar that allows companies to access the page 

### LINKED ISSUE #74: Add MyOffers page
Create page for companies to see their offers. This page should be available to users that are logged in

monkin77:
Closes #74


src/components/HomePage/SearchResultsArea/Offer/Offer.js
```diff
@@ -1,10 +1,11 @@
...
+    constructor({ id, position, company, location, date, description, fields }) {
         this.id = id;
         this.position = position;
         this.company = company;
         this.location = location;
         this.date = date;
         this.description = description;
+        this.fields = fields;
     }
 }
```
imnotteixeira, line 11: Don't edit this file in this PR, it's already done in #78, and it is more complete than this

src/components/Navbar/navbarStyles.js
```diff
@@ -11,9 +11,9 @@ export default makeStyles((theme) => ({
     homePageLink: ({ isMobile }) => ({
         marginLeft: theme.spacing(isMobile ? 0 : 3),
     }),
-    linkStyle: ({ forceDesktopLayout }) => ({
+    linkStyle: ({ isMobile }) => ({
         textDecoration: "none",
-        color: !forceDesktopLayout ? theme.palette.secondary.main : "white",
+        color: isMobile  ? theme.palette.secondary.main : "white",
```
imnotteixeira, line 16: This `isMobile` check has problems as well as the `forceDesktopLayout`. We'll need a combination of both.

Check for example the Not Found page (by going to some random route) in desktop and mobile. Since it forces the desktop layout in mobile, the color will be wrong there. That would also happen in the homepage itself, but doesn't render a homepage button, so it is not as problematic.

I think the condition we need is `isMobile && !forceDesktopLayout`, but please review and test it in multiple pages in the different layouts

src/pages/MyOffersPage.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import { Paper } from "@material-ui/core";
+import CenteredComponent from "../components/HomePage/CenteredComponent";
+import { MainMask } from "../components/HomePage/MainMask";
+
+export const MyOffersPage = () => (
+    <>
+        <MainMask />
+        <CenteredComponent>
+            <Paper style={{ width: "60%", padding: "24px 72px", boxSizing: "content-box" }}>
```
imnotteixeira, line 10: We should use a `<Card>` instead of a `<Paper>`. Also, it is a bad practice to have a style object applied directly like so (https://marmelab.com/blog/2017/02/06/react-is-slow-react-is-fast.html#beware-of-object-literals-in-jsx) (I know it is being done in other places, but it should be fixed as well!!!)

Please use the `useStyles` approach.

src/components/Navbar/UserMenu.js
```diff
@@ -100,10 +100,8 @@ const UserMenuContent = React.forwardRef(({ open, isMobile = false, sessionData,
                 autoFocusItem={open}
                 id="menu-list-grow"
             >
-                {/*
-                <MenuItem button disableTouchRipple onClick={() => {}}>My Offers</MenuItem>     // These options need to be implemented
+                <MenuItem button component={Link} to="/company/offers/manage" disableTouchRipple>My Offers</MenuItem>
```
imnotteixeira, line 103: This should only show up to company users, similar to what happens with AdminMenuOptions

src/pages/MyOffersPage.js
```diff
@@ -0,0 +1,17 @@
...
+                <h1>My Offers Page</h1>
```
imnotteixeira, line 11: This title will be in conflict with the PageLayout title in mobile. I'd say you wait for the work I'm doing in the Reactive Pattern (I'm refactoring all pages that have desktop/mobile versions) and deal with this logic after that.

imnotteixeira (changes requested):

imnotteixeira:
It looks ok, but I'll wait for #100 to go through before merging this one, so that we don't have to refactor yet another component

src/AppRouter.js
```diff
@@ -73,6 +75,21 @@ const AppRouter = () => (
                     <ApplicationsReviewPage />
                 </PageLayout>
             </ProtectedRoute>
+            <ProtectedRoute
+                exact
+                path="/company/offers/manage"
+                unauthorizedRedirectPath="/"
+                unauthorizedRedirectMessage="You are not allowed to access the my offers page"
```
imnotteixeira, line 82:
```suggestion
                unauthorizedRedirectMessage="You are not allowed to access the 'My Offers' page"
```

src/AppRouter.js
```diff
@@ -73,6 +75,21 @@ const AppRouter = () => (
...
+                authorize={(user) => !!(user?.company)}
+            >
+                <PageLayout
```
imnotteixeira, line 85: PageLayout needs a unique `key` to work properly, as per the comment on the beginning of this file.

Use the route path for example

src/pages/MyOffersPage.js
```diff
@@ -0,0 +1,29 @@
...
+    const classes = useStyles();
+    return (
+        <>
+            <MainMask />
+            <CenteredComponent>
+                <Card className={classes.cardContainer}>
+                    <h1>My Offers Page</h1>
+                </Card>
+            </CenteredComponent>
+        </>
```
imnotteixeira, line 24: By using a PageLayout, you only need to render the content of the Card, ideally in a `<CardContent/>`
Also, PageLayout already handles the card style, so you won't need the styles object probably

```suggestion
<CardContent>
        <h1>My Offers Page</>
</CardContent>
```

imnotteixeira (changes requested):

src/pages/MyOffersPage.js
imnotteixeira: Also, don't bother having a title for this, since PageLayout will put the title automatically for you!

You can have the content straight away

src/AppRouter.js
monkin77: I believe that is already implemented in line 86

src/pages/MyOffersPage.js
monkin77: Got it. For now I left with the header since the content will be on another pr

imnotteixeira (approved):

PR #77, opened 2021-03-01
Feature/beautify error page

### LINKED ISSUE #58: Beautify Error/Not Found pages
The current error page looks like this:

image

The current 404/not found page looks like this:

image

Make them use the "theme" of the rest of application by using the `<MainMask/>` and `<CenteredComponent/>` with a Paper with some message there. It should work well on mobile as well. On mobile, the main mask should be "cut" and it will only be a red rectangle.

The route->page mapping is done on the AppRouter.js. `/error` renders an ErrorPage.js, and an unknown resource renders a NotFound.js, both present in `src/pages`

TiagooGomess:
### Current design:
Not found page:
Screenshot from 2021-03-01 17-15-37

Error page:
Screenshot from 2021-03-01 17-23-30

Closes #58

### DIFF
--- a/src/components/Errors/ErrorComponent.js
+++ b/src/components/Errors/ErrorComponent.js
@@ -6,17 +6,16 @@ import { MainMask } from "../HomePage/MainMask";
 import CenteredComponent from "../HomePage/CenteredComponent";
 
 const useStyles = makeStyles((theme) => ({
-    card: {
-        textAlign: "center",
-    },
     title: {
         marginTop: theme.spacing(1),
+        marginLeft: theme.spacing(1),
     },
     content: {
-        textAlign: "center",
+        marginLeft: theme.spacing(1),
     },
     secondText: {
         marginBottom: theme.spacing(1),
+        marginLeft: theme.spacing(1),
     },
 }));
 
@@ -27,12 +26,12 @@ const EmailInfo = ({ email }) => {
             <Typography variant="body1" className={classes.secondText}>
                 Please try again later, and if the problem persists, contact us at
                 {" "}
-                <Link color="secondary" href="mailto:ni@aefeup.pt">
+                <Link color="secondary" href={`mailto:${email}`}>
                     {email}
                 </Link>
             </Typography>
         );
-    return <Typography />;
+    return null;
 };
 
 EmailInfo.propTypes = {
--- a/src/components/Errors/ErrorComponent.js
+++ b/src/components/Errors/ErrorComponent.js
@@ -5,24 +5,21 @@ import { Paper, Typography } from "@material-ui/core";
 import { MainMask } from "../HomePage/MainMask";
 import CenteredComponent from "../HomePage/CenteredComponent";
 
-const ErrorComponent = ({ Message }) => (
-
+const ErrorComponent = ({ message }) => (
     <React.Fragment>
-        <MainMask/>
+        <MainMask />
         <CenteredComponent>
-
             <Paper elevation={8} style={{ padding: "24px" }}>
                 <Typography align="center">
-                    {Message}
+                    {message}
                 </Typography>
             </Paper>
-
         </CenteredComponent>
     </React.Fragment>
 );
 
 ErrorComponent.propTypes = {
-    Message: PropTypes.string,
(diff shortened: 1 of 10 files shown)


src/pages/ErrorPage.js
```diff
@@ -1,7 +1,9 @@
 import React from "react";
 
+import ErrorComponent from "../components/Errors/ErrorComponent";
+
 const ErrorPage = () => (
-    <p>Something is wrong in the backstage. :(</p>
+    <ErrorComponent Message={"Something is wrong in the backstage. :("}/>
```
imnotteixeira, line 6: The old message was just a placeholder. Please change to something more formal like "Something unexpected prevented us from fulfilling your request. Please try again later, and if the problem persists, contact us at ni@aefeup.pt (make this an actual `mailto` link)"

src/pages/NotFound.js
```diff
@@ -1,12 +1,12 @@
...
+import ErrorComponent from "../components/Errors/ErrorComponent";
+
 class NotFound extends Component {
 
     render() {
         return (
-            <div>
-                Oh no! Someone took this page away!
-            </div>
+            <ErrorComponent Message={"Sorry, no page could be found at this address (404)"}/>
```
imnotteixeira, line 9: If you add a title prop to the error component, you can keep it generic and pass a different title and message for 404 and unexpected errors

src/pages/NotFound.spec.js
```diff
@@ -1,8 +1,18 @@
...
+import { renderWithTheme } from "../test-utils";
+
+const theme = createMuiTheme({});
 
 describe("NotFound page", () => {
     it("should render successfully", () => {
-        mount(<NotFound />);
+        const errorMessage = "Sorry, no page could be found at this address (404)";
+        const wrapper = renderWithTheme(
+            <ErrorComponent Message={errorMessage}/>,
```
imnotteixeira, line 13: In order to test the Not Found page, you should not render its contents, but the page itself, and assert that the contents are as expected. Change the component to just the <NotFound> and assert that the error message is whatever you'll put on the component above. If you add the title, add here the assertion that the title is present as well.

Finally, add a similar test for the error page as well

imnotteixeira (changes requested):

TiagooGomess:
## Page not found:
Screenshot from 2021-03-07 16-41-37

## Error page:
Screenshot from 2021-03-07 16-43-51

imnotteixeira (changes requested):
On the mobile version, the homepage icon (button in the navbar) should be white. That happens because it is not taking into account the forceDesktopLayout. Only when forceDesktopLayout=false it should be the dark red, since the background will be the white from the dialog.

Regarding the ErrorComponent, the text should be left aligned (both title and content)

Finally, you should unstage the package*.json -related changes and drop irrelevant commits, to avoid conflicts in the future, but regarding this we can sync "offline"

TiagooGomess:
## Desktop view:
Screenshot from 2021-03-14 14-50-38

## Mobile view:
Screenshot from 2021-03-14 14-50-50

src/components/Errors/ErrorComponent.js
```diff
@@ -0,0 +1,67 @@
...
+}));
+
+const EmailInfo = ({ email }) => {
+    const classes = useStyles();
+    if (email)
+        return (
+            <Typography variant="body1" className={classes.secondText}>
+                Please try again later, and if the problem persists, contact us at
+                {" "}
+                <Link color="secondary" href="mailto:ni@aefeup.pt">
```
imnotteixeira, line 29:
```suggestion
                <Link color="secondary" href={`mailto:${email}`}>
```

src/components/Errors/ErrorComponent.js
```diff
@@ -0,0 +1,67 @@
...
+                    {email}
+                </Link>
+            </Typography>
+        );
+    return <Typography />;
```
imnotteixeira, line 34: return null instead

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #78, opened 2021-03-04
Add base offer searching functionality with filters

imnotteixeira:
Even though the title says 'base offer searching', this is not basic search. It includes advanced search with filters and full text search.

The base meaning here is that it needs to be extended in the future to have some infinite loading features, including result list virtualization (#79). However, for a small number of offers, this works as a PoC. Also, in the future we can add more filters to advanced search, as long as the backend supports them.


src/components/HomePage/SearchArea/AdvancedSearch/FieldOptions.js
```diff
@@ -1,9 +1,9 @@
...
-    { label: "Front-End", value: "FRONT_END" },
-    { label: "Dev-Ops", value: "DEVOPS" },
-    { label: "Machine Learning", value: "ML" },
-    { label: "Computer Vision", value: "COMPUTER_VISION" },
-]);
+export default Object.freeze({
+    "BACKEND": "Back-End",
+    "FRONTEND": "Front-End",
+    "DEVOPS": "Dev-Ops",
+    "ML": "Machine Learning",
```
monkin77, line 7: To match the backend, some keys should be changed. Here should be: "MACHINE LEARNING".

src/components/HomePage/SearchArea/AdvancedSearch/FieldOptions.js
```diff
@@ -1,9 +1,9 @@
...
+    "COMPUTER_VISION": "Computer Vision",
```
monkin77, line 8: There is no COMPUTER_VISION in the backend aswell, only an "OTHER" field, so I am not sure if this is intended. To check this, the file from the backend is: "FieldTypes.js"

src/components/HomePage/SearchArea/AdvancedSearch/TechOptions.js
```diff
@@ -1,9 +1,10 @@
...
 
-export default Object.freeze([
-    { label: "React.js", value: "REACT_JS" },
-    { label: "Node.js", value: "NODE_JS" },
-    { label: "Python", value: "PYTHON" },
-    { label: "Java", value: "JAVA" },
-    { label: "C++", value: "C++" },
-]);
+export default Object.freeze({
+    "REACT_JS": "React.js",
```
monkin77, line 4: These keys are also different to the ones on "TechnologyTypes.js" from the backend. I've noticed that in the backend the Technology types are not all capital letters, while the Field types are all Capital letters. I am not sure if it is intended.

src/components/HomePage/SearchArea/JobTypes.js
```diff
@@ -12,11 +12,11 @@ const JobTypes = Object.freeze([
         label: "Curricular Internship",
     },
     {
-        value: "PART_TIME",
+        value: "PART-TIME",
```
monkin77, line 15: The keys in this file are also different from the ones on "JobTypes.js" from the backend. Also, Is there a reason why this file is not in the same location as the TechOptions / FieldOptions? also I believe its name should be "JobOptions.js" just to be coherent with the others.

monkin77 (changes requested):
Overall it works fine, but there are some minor issues that make some filters not work

src/components/HomePage/SearchArea/AdvancedSearch/TechOptions.js
imnotteixeira: This is not in the best state right now, but this issue will attempt at making it all better NIAEFEUP/nijobs-be#103 by making both projects sync on these values

monkin77 (approved):

PR #80, opened 2021-03-05
Fix useSession reset function to actually reset session data

### LINKED ISSUE #76: [Bug] Logout button does not update session details
To Reproduce:

1. Login with some account
2. Logout
3. Notice that the UI still appears as if the user is logged in (though it is not, as you can verify by refreshing the page)

The behavior should be such that when the user logs out, the UI updates so that it shows that the user is logged out.

I believe this to be a problem with the useSession custom hook, but further investigation is necessary.

imnotteixeira:
Closes #76 

The `mutate` calls were not correctly defined, since they were relying on the bound version (currently implemented in this PR), but they were using the general version (imported from swr directly). Now it should be working as expected, as you can verify, by using your own backend and set it in the devtools in netlify preview.

### DIFF
--- a/.env
+++ b/.env
@@ -1,7 +1,7 @@
 HOST_PORT=3000
 # PUBLIC_URL=https://ni.fe.up.pt/st4g1ng/nijobs/
 # REACT_APP_BASE_ROUTE=test/test123
-REACT_APP_API_HOSTNAME="http://localhost:8087"
+REACT_APP_API_HOSTNAME="https://localhost:8087"
 
 # Uncomment to disable devtools
 # REACT_APP_ALLOW_DEV_TOOLS=false
--- a/README.md
+++ b/README.md
@@ -62,22 +62,24 @@ For it to work, you must serve a backend through some server accessible on the i
 After starting your server on localhost, you can create a tunnel from that localhost server to the internet with ngrok with the following command:
 
 ```bash
-ngrok http <backend port, usually 8087>
+ngrok http https://localhost:8087 --region eu # chage port if not using default 8087
 ```
 
 That will give you two hosts, one for `http`, another for `https`. Use the `https` one in the NIJobs Devtools.
 
-Remember that the backend server must allow the host making the requests (the Netlify origin (i.e. `https://deploy-preview-66--nijobs.netlify.app/`), or your localhost (i.e. `http://localhost:3000`), depending on the use-case).
+Remember that the backend server must allow the host making the requests (the Netlify origin (i.e. `https://deploy-preview-66--nijobs.netlify.app/`), or your localhost (i.e. `https://localhost:3000`), depending on the use-case).
+
+Also, since the development backend uses a self-signed certificate, your browser might block it by default. To fix this, simply visit an endpoint and allow it (e.g. https://localhost:8087).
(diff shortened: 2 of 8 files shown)


monkin77 (approved):

monkin77:
When a user logs out, he may get stuck on a page that he has no access to, such as the CompanyApplication page. Not sure if it is the desired behaviour

imnotteixeira:
> 
> 
> When a user logs out, he may get stuck on a page that he has no access to, such as the CompanyApplication page. Not sure if it is the desired behaviour

It should work correctly, since the hook updates the session state, and the ProtectedRoute re-renders, but currently it's in an infinite loop, I will investigate further

imnotteixeira:
> > When a user logs out, he may get stuck on a page that he has no access to, such as the CompanyApplication page. Not sure if it is the desired behaviour
> 
> It should work correctly, since the hook updates the session state, and the ProtectedRoute re-renders, but currently it's in an infinite loop, I will investigate further

After some investigation, I think I found the issue. Basically after the logout (resetting the session details by calling reset() from useSession hook - which in turn calls mutate(null) on the SWR state) the session data would become `null` as expected.

However, by looking at the condition in the ProtectedRoute: https://github.com/NIAEFEUP/nijobs-fe/blob/b2935f137928cba136528515ce143dd5b861d10d/src/utils/index.js#L163

we see that by having data === null and no error will trigger an infinite Loading state. 

So I realized what the problem was: After logging out, the session data must be `revalidated` (calling revalidate on the SWR state), which will trigger a call to /auth/me which will in turn return a 401 response since the user is logged out). This will populate the `error` object, and it will now have data === null, but with an error, which will trigger the Redirect, as expected :)

Pushed a commit with this fix, please check again!

monkin77:
> > > When a user logs out, he may get stuck on a page that he has no access to, such as the CompanyApplication page. Not sure if it is the desired behaviour
> > 
> > 
> > It should work correctly, since the hook updates the session state, and the ProtectedRoute re-renders, but currently it's in an infinite loop, I will investigate further
> 
> After some investigation, I think I found the issue. Basically after the logout (resetting the session details by calling reset() from useSession hook - which in turn calls mutate(null) on the SWR state) the session data would become `null` as expected.
> 
> However, by looking at the condition in the ProtectedRoute:
> 
> https://github.com/NIAEFEUP/nijobs-fe/blob/b2935f137928cba136528515ce143dd5b861d10d/src/utils/index.js#L163-L164
> 
> we see that by having data === null and no error will trigger an infinite Loading state.
> 
> So I realized what the problem was: After logging out, the session data must be `revalidated` (calling revalidate on the SWR state), which will trigger a call to /auth/me which will in turn return a 401 response since the user is logged out). This will populate the `error` object, and it will now have data === null, but with an error, which will trigger the Redirect, as expected :)
> 
> Pushed a commit with this fix, please check again!

I confirmed it, it is now working fine :)

imnotteixeira:
image

This is why it is not working in chrome, specifically.

This only affects dev and chrome specifically, since in firefox the cookies are not blocked. So I believe we can merge this safely.

I will create another issue describing the problem and how to fix.

PR #91, opened 2021-03-21
Dynamic website title based on current page

### LINKED ISSUE #82: Dynamic Website title based on current page
Currently the title is hardcoded to "NIJobs" on `public/index.html`. This is fine, however good practices dictate that the title should update based on the current page. Currently, we can use the PageLayout title prop to derive the page title so we can have something like "<page title> - NIJobs"

To do this we can use the Document Title API https://developer.mozilla.org/en-US/docs/Web/API/Document/title, and call it on mount to change the title inside PageLayout

```javascript
//PageLayout

.....

useEffect(() => {
    document.title = `${pageTitle} - NIJobs`
}, []);
```

TiagooGomess:
Closes #82

### DIFF
--- a/src/components/PageLayout.js
+++ b/src/components/PageLayout.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import ContactSection from "./HomePage/ContactSection";
 import PropTypes from "prop-types";
 import Navbar from "./Navbar/index";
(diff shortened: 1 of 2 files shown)


src/components/PageLayout.js
```diff
@@ -7,6 +7,10 @@ import { Dialog } from "@material-ui/core";
 
 const PageLayout = ({ children, showHomePageLink = true, pageTitle, forceDesktopLayout = false }) => {
 
+    useEffect(() => {
+        document.title = pageTitle !== undefined ? `${pageTitle} - NIJobs` : "NIJobs";
```
imnotteixeira, line 11: This would fail if pageTitle = null, for example, since `null !== undefined`. You can just use the boolean version of the variable on the check
```suggestion
        document.title = pageTitle ? `${pageTitle} - NIJobs` : "NIJobs";
```

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #92, opened 2021-03-25
Feature/my offers table

monkin77:
This pr will be to refactor the Table Components that are used on the ApplicationReviewPage so that we can use them with any page. Furthermore, it is also to create a first look of the CompanyOffersManagementPage, where it will display the offers a company has created, for now mock offers.


monkin77:
I've removed the route to the CompanyOffersManagementPage since it is added in the myOffersPage branch

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -0,0 +1,125 @@
...
+    };
+
+    RowComponent.propTypes = {
+        rowKey: PropTypes.string.isRequired,
+        labelId: PropTypes.string.isRequired,
+    };
+
+    return (
+        <div>
+            <Card>
```
imnotteixeira, line 92: After the refactor from PageLayout, you should remove this Card, since it will already be rendered above

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -0,0 +1,125 @@
...
+
+
+const CompanyOffersManagementWidget = () => {
+    const [offers, setOffers] = useState(null);
+    const [isLoading, setIsLoading] = useState(true);
+    const [error, setError] = useState(null);
+
+    useEffect(() => {
+        fetchCompanyOffers().then((offers) => {
+            /* TODO: SHOULD NOT RUN WHEN COMPONENT IS UNMOUNTED: CANCEL PROMISE */
```
imnotteixeira, line 59: After #96, you'll be able to cancel this promise on unmount. If you review it, we can merge :eyes:

imnotteixeira (changes requested):
I won't review this PR fully, since it is still missing some parts on the offers management feature.

Since this is a refactor to the tables, if you create a PR with only that part (remove anything related to company offers management, but leave any necessary change to applications review for example) I can review.

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -0,0 +1,33 @@
...
+        renderWithStore(
+            <MuiPickersUtilsProvider utils={DateFnsUtils}>
+                <CompanyOffersManagementWidget />
+            </MuiPickersUtilsProvider>
+        );
+
+        // Use getBy by default when element should be available
+        expect(screen.getByText("Loading...")).toBeInTheDocument();
+
+        expect(screen.getByRole("heading")).toBeInTheDocument();
```
imnotteixeira, line 21: Avoid using getByRole queries as they are too slow. besides this query is too generic, I believe.

If the heading you are referring to here is the same that shows "Loading...", you can remove this line, since you are testing that above

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+        // Use queryBy When asserting for missing element
+        expect(screen.queryByText("Offers Management")).toBeNull();
+
+        screen.debug();
```
imnotteixeira, line 26: remove this

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -0,0 +1,33 @@
...
+            </MuiPickersUtilsProvider>
+        );
+
+        // Use getBy by default when element should be available
+        expect(screen.getByText("Loading...")).toBeInTheDocument();
+
+        expect(screen.getByRole("heading")).toBeInTheDocument();
+
+        // Use queryBy When asserting for missing element
+        expect(screen.queryByText("Offers Management")).toBeNull();
```
imnotteixeira, line 24:
```suggestion
        expect(screen.queryByText("Offers Management")).not.toBeInTheDocument();
```

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+import CompanyOffersManagementWidget from "./CompanyOffersManagementWidget";
+import { renderWithStore } from "../../../../test-utils";
+import { MuiPickersUtilsProvider } from "@material-ui/pickers";
+import DateFnsUtils from "@date-io/date-fns";
+
+
+describe("App", () => {
+    test("Renders Component", async () => {
+        renderWithStore(
```
imnotteixeira, line 12: you should wrap this in an act() call, since the render of the component itself will cause a self-update (due to the fetch of offers). In order to test correctly the different states, you should mock the service you are using so that it does not return (to test loading) in a test case, and then another where the service returns something

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+        expect(screen.getByRole("heading")).toBeInTheDocument();
+
+        // Use queryBy When asserting for missing element
+        expect(screen.queryByText("Offers Management")).toBeNull();
+
+        screen.debug();
+
+        // Use find by for elements that aren't there initially, but will be eventually
+        expect(await screen.findByText("Offers Management")).toBeInTheDocument();
```
imnotteixeira, line 29: I believe that if you use act() on the render above, you won't need the await find* query, please try it with a normal get and see if it works, as it will be faster

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+        screen.debug();
```
imnotteixeira, line 31: remove

src/components/Review/Applications/Actions.js
```diff
@@ -116,14 +116,14 @@ const RowActionsContainer = ({
...
-    handleApprove: PropTypes.func.isRequired,
-    executingAction: PropTypes.bool.isRequired,
+    isCollapseOpen: PropTypes.bool,     /* Removed isRequired from all props */
+    toggleCollapse: PropTypes.func,
+    actionToConfirm: PropTypes.func,
+    confirmAction: PropTypes.func,
+    cancelAction: PropTypes.func,
+    handleAction: PropTypes.func,
+    handleApprove: PropTypes.func,
+    executingAction: PropTypes.bool,
```
imnotteixeira, line 126: These props refer to the actions of RowACtions from Review/Applications. They should not be reused as is.

If you want, you can create a more generic "RowActions*" component that both this and the new offers table use to create their custom actions, with their custom propTypes rules. Otherwise, create a similar component for the offers table feature.

Bottom line, these props must be required for now, don't change them just for reusability

imnotteixeira (changes requested):

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
monkin77: Yes, I just left it there as I was testing react-native-testing-library functionalities. I will try to make more thorough tests

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
monkin77: I've tried to use only with the act and it was not working. From what I've searched, since render() is a synchronous function, it only flushes synchronous state updates and that was not the case.

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
monkin77: I've wrapped it around the act(). However, I wasn't able to mock the service, since on this branch the service itself is returning a MOCK_OFFERS, so it doesn't use fetch. 
I'm not sure if that is the only way, but from what I searched, to mock the service I should mock the fetch call. So I left a comment and maybe I complete the test on the branch where the service doesn't return a mock?

src/utils/Table/TableContent.js
```diff
@@ -67,7 +67,7 @@ TableRow.propTypes = {
     rowProps: PropTypes.object.isRequired,
     handleSelect: PropTypes.func.isRequired,
     isRowSelected: PropTypes.func.isRequired,
-    RowActions: PropTypes.elementType.isRequired,
+    RowActions: PropTypes.elementType,
```
monkin77, line 70: I changed this since we could have tables without RowActions. Is that alright? or should we enforce it?

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
imnotteixeira: render is synchronous, but act is not :P You can await act(()=>{render...}), and from what I know, it will execute all the async code at once ( the batch changes from useEffects, setStates etc as if they were running in a browser (where they happen in batches, not synchronously)

src/utils/Table/TableContent.js
imnotteixeira: For that case you can remove it, yes.

monkin77:
The tests should be working fine now, however, there are a lot of warnings while running them, such as this one
image
I'm not sure what's wrong, should they be fixed?

imnotteixeira:
> 
> 
> The tests should be working fine now, however, there are some warnings while running them, such as this one
> image
> I'm not sure what's wrong, should they be fixed?

Whenever a component updates ( in this case the component updates itself vie useEffect I believe), you should wrap its render call on the test with an act() so that the update works as expected. Otherwise, it may fail to simulate the correct behavior.

Maybe read this: https://davidwcai.medium.com/react-testing-library-and-the-not-wrapped-in-act-errors-491a5629193b

monkin77:
> > The tests should be working fine now, however, there are some warnings while running them, such as this one
> > image
> > I'm not sure what's wrong, should they be fixed?
> 
> Whenever a component updates ( in this case the component updates itself via useEffect I believe), you should wrap its render call on the test with an act() so that the update works as expected. Otherwise, it may fail to simulate the correct behavior.
> 
> Maybe read this: https://davidwcai.medium.com/react-testing-library-and-the-not-wrapped-in-act-errors-491a5629193b

Thanks, I've read it and applied one of its solutions

monkin77:
- [ ] Assume Offer.js is like in the other branch (Change the reference to the variable like company.name) See in the View Offer Page branch

monkin77:
@imnotteixeira I'm not sure if this was talked about already, but what columns should the table have?
Should it include the owner of the offer as well?
image

imnotteixeira:
> @imnotteixeira I'm not sure if this was talked about already, but what columns should the table have?
> Should it include the owner of the offer as well?
> image

I don't think so, at least in the context of the My Offers page, since the viewer will be the owner. When we have the admin view of the existing offers, it might be appropriate.

imnotteixeira:
I think the current schema is ok, but why are the dates only showing mm-yyyy? If the companies can select the day, it should be shown as well

monkin77:
> I think the current schema is ok, but why are the dates only showing mm-yyyy? If the companies can select the day, it should be shown as well

Yea, that was only the MOCK_OFFERS date that was wrong

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -0,0 +1,120 @@
...
+    } else {
+        return null;
+    }
+};
+
+const generateRow = ({ title, location, publishDate, publishEndDate }) => ({
+    fields: {
+        title: { value: title, align: "left" },
+        publishStartDate: { value: publishDate },
+        publishEndDate: { value: publishEndDate },
```
imnotteixeira, line 31:
```suggestion
        publishStartDate: { value: format(publishDate), "yyyy-MM-dd") },
        publishEndDate: { value: format(publishEndDate), "yyyy-MM-dd") },
```

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #94, opened 2021-03-31
Add home page product description

### LINKED ISSUE #64: Add information about the tool to the Home Page
imagem

Mocks are TBD but we should have something like 3 columns with icons stating our motivation/goals, and how companies can use this and how to join/login with links

pedrojfs17:
Closes #64


src/components/HomePage/ProductPlacementArea/productDescriptionStyles.js
```diff
@@ -0,0 +1,45 @@
...
+        "& > p:not(:first-child)": {
+            margin: "1em 0 0.5em 0",
+        },
+    },
+    productDescriptionAnchor: {
+        width: "100%",
+        height: "100%",
+        "&:hover": {
+            textDecoration: "none",
+        },
```
imnotteixeira, line 42: Instead of doing most of this "by hand", which is harder to maintain, you should refactor and use [Grid](https://material-ui.com/components/grid/). ProductDescription would be a grid container, and each column would be a grid item with xs=12 (this makes each column take the full space in mobile (xs)) and md=6 (taking only half space in md size and above).

You might want to experiment with the thresholds to see what fits best (using `sm` instead of `md`, for example, but the base is this one) Then you'll only need to give the main component (product description) some styles to have bigger side margins to keep it similar to what it is now and center it (I don't know if it will center automatically) -> you can just wrap the grid container in another flex div, and add `justify-content: center`, without changing `flex-direction`

imnotteixeira (changes requested):

imnotteixeira:
Please change the color of the svgs to `#999`, you can do that by adding the fill property to the svg element.

Additionally, make sure that the login/join us buttons are side by side (disabling 100% width might be enough)

Also, I think it would make sense to add a subtitle to each column with "Are you a Student?" / "Are you a Company?" with an `h6` variant of Typography.

The content still needs to change, but we can think about it later.

src/components/HomePage/ProductPlacementArea/ProductDescription.js
```diff
@@ -13,28 +20,56 @@ import useSession from "../../../hooks/useSession";
...
+                            onClick={toggleLoginModal}
+                        >
+                        Login
+                        </Button>
+                        <Button
+                            variant="text"
+                            color="primary"
+                        >
+                            <RouterLink className={classes.productDescriptionAnchor} to="/apply/company">Join Us</RouterLink>
+                        </Button>
```
imnotteixeira, line 66: You can use the `component` prop to change the displayed component, instead of using children. https://material-ui.com/guides/composition/#button

That might be helpful for avoiding manual styling as well.

```suggestion
                        <Button
                            color="primary"
                            component={RouterLink}
                            to="/apply/company"
                        >
                            Join Us
                        </Button>
```

imnotteixeira (changes requested):

src/components/HomePage/ProductPlacementArea/ProductDescription.js
pedrojfs17: When making these changes, the page sends this warnings:
image
image
image

imnotteixeira: As discussed "offline", you need to make RouterLink a "refForwarder". Coincidentally, that pattern is already done inside RouterLink, but not on the required level. You need to copy that "forwardRef" call to the RouterLink function itself (not inside - and do not remove the existing forwardRef call!) and pass the ref to the Link component

src/components/HomePage/ProductPlacementArea/ProductDescription.js
```diff
@@ -4,37 +4,103 @@ import { RouterLink } from "../../../utils";
...
+                        <CompanySvg fill="#999" className={classes.productDescriptionSVG} />
+                    </Grid>
+                    <Grid item>
+                        <Typography variant="h6">
+                            Are you a company?
+                        </Typography>
+                    </Grid>
+                    <Grid item>
+                        <Typography>
+                            As a Company, you can advertise yout job opportunities to students!
```
imnotteixeira, line 59:
```suggestion
                            As a Company, you can publish your job or internship opportunities to students!
```

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #96, opened 2021-04-06
Allow promises from services to be cancellable

imnotteixeira:
This PR will allow async work done by services to be cancelable from React components (mostly useful when unmounting components, in order to cancel any work on unmounted components)

Additionally, since we are using fetch, we can actually cancel the requests with the AbortableController API (https://developer.mozilla.org/en-US/docs/Web/API/AbortController) - This would be done on the `onCancel` callback of the promises.


monkin77 (changes requested):
Everything seemed to be working well. The only thing missing is to rebase from develop to get the current changes.

TiagooGomess (approved):

monkin77 (approved):

PR #97, opened 2021-04-07
Rules page

AvilaAndre:
This is the rules page, it contains accordions with the rules, I didn't wrote them but they can be written on the rules.json file, there will be as many rule accordions as there are rules on the rules.json file.


src/pages/RulesPage.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import RulesComponent from "../components/RulesPage";
+
+export const RulesPage = () => (
+    <div>
```
imnotteixeira, line 5:
```suggestion
    <div className={classes.rulesCard}>
```

src/pages/RulesPage.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import RulesComponent from "../components/RulesPage";
+
```
imnotteixeira, line 3:
```js
const useStyles = (isMobile) => makeStyles((theme) => ({
    rulesCard: {
        padding: isMobile ? theme.spacing(0, 1) : theme.spacing(10),
        maxWidth: !isMobile && "80%"
    },
}));
```

imnotteixeira (changes requested):
To solve the problem with the width and adjust the padding to make it consistent with other pages

src/pages/RulesPage.js
DoStini: Is this solved with the rebase?

src/components/RulesPage/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+const useStyles = makeStyles((theme) => ({
+    ruleCard: {
+        marginBottom: theme.spacing(2),
+    },
+}));
+
+// Writing on rules.json adds "Accordions" with json's data. You may add a title, a description,
+// a link and a linkName, if link is different than "", a button appears, this button will have, linkName's
+// text, clickign it will take you to a page inside ni-jobs as the link is just the path part of router-dom's
```
imnotteixeira, line 23: fix typo

imnotteixeira: also "NIJobs" :roll_eyes:

imnotteixeira (changes requested):

src/pages/RulesPage.js
imnotteixeira: Maybe (?)

I looked at the preview, and the only thing I think is missing is a padding with `spacing(4)`. So maybe add that class with it anyway.

src/utils/Constants.js
```diff
@@ -0,0 +1,3 @@
+export default {
+    EMAIL: "nijobs@aefeup.pt",
```
imnotteixeira, line 2:
```suggestion
    CONTACT_US_EMAIL: "nijobs@aefeup.pt",
```

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #101, opened 2021-04-11
Fix Application Layout

### LINKED ISSUE #100: Allow PageLayout to have a layout selection
As discussed in a meeting, the application layout should be consistent, and there are a few scenarios that need to be supported, as shown in the amazing sketches below:

There should be 3+1 Layout types:
* None: No specific layout is enforced, but navbar, main mask (the red thing on the background) and the footer are still there
* Desktop, no scroll: single centered component with navbar, main mask and footer
* Desktop, scroll: Single initially centered component that grows downwards, allowing page scroll, with navabar, main mask and footer as well
* Mobile: Full page card with homepage button and page title on top, and footer on bottom. Should apply automatically when in mobile, or when a certain criteria occurs (given to PageLayout)

Each page defines its layout behavior by passing props to PageLayout.

imnotteixeira:
Closes #100 

This fixes the layout consistency problems and guarantees that pages will look similar to each other.

Currently, there will only be 3 types of layouts (`None`, `Desktop` and `Mobile`) as described in the issue. However we can easily add more if required (maybe admin/back-office pages could be different?)

Pages to refactor:

- [x] Homepage
- [x] Applications Review
- [x] Company Application (Also refactored it to use the Reactive pattern in order not to lose data between desktop/mobile changes)
- [x] Error Page
- [x] Not Found


monkin77 (approved):
Worked flawlessly for me

PR #103, opened 2021-04-21
Finish Company Registrations

### LINKED ISSUE #53: First login experience
When a user first logs in, it's necessary to complete some registration details such as the company bio and logo.

Create an experience that only happens on first log in (or while the registration is not complete) that allows the user to complete a set of steps, in order to complete the registration

1. Add a company logo
1. Add a company bio
1. Add company contacts
1. Upon completion, provide a CTA (Call-to-Action) for them to create offers right away

In order for it not to be very invasive, we can have this as a banner on the top of the page, and when trying to create an offer, do not allow it while the company registration is not complete.

imnotteixeira:
Closes #53 

* Shows a notification for companies that haven't finished their registration on login with a link for the page at `/company/application/finish`
* That page has a "wizzard-like" form where companies follow 3 steps in order to finish their registration
  * Upload a logo
  * Add a Bio
  * Add company contacts

https://user-images.githubusercontent.com/28157246/117086684-292b8a00-ad45-11eb-87c0-96b2eeff2d4b.mp4

### DIFF
--- a/src/components/ApplicationMessages/index.js
+++ b/src/components/ApplicationMessages/index.js
@@ -1,0 +1,50 @@
+import React, { useEffect, useState } from "react";
+import PropTypes from "prop-types";
+import { addSnackbar } from "../../actions/notificationActions";
+import useSession from "../../hooks/useSession";
+import { connect } from "react-redux";
+import Notification from "../Notifications/Notification";
+import { useHistory } from "react-router";
+
+const ApplicationMessageNotifier = ({ addSnackbar }) => {
+
+    const [hasShownCompleteRegistrationMessage, setHasShownCompleteRegistrationMessage] = useState(false);
+
+    const history = useHistory();
+
+
+    const { data, isLoggedIn } = useSession();
+    useEffect(() => {
+        if (isLoggedIn && !data.company?.hasFinishedRegistration && !hasShownCompleteRegistrationMessage) {
+            setHasShownCompleteRegistrationMessage(true);
+            addSnackbar({
+                message: "In order to fully use NIJobs, you still need to finish your registration.",
+                options: {
+                    // eslint-disable-next-line react/display-name
+                    content: (key, message) =>
+                        <Notification
+                            message={message}
+                            actionText="Show me"
+                            actionHandler={() => {
+                                history.push("/company/registration/finish");
+                            }}
+                        />,
+                },
+            });
+        }
+        if (hasShownCompleteRegistrationMessage && !isLoggedIn) setHasShownCompleteRegistrationMessage(false);
+    }, [addSnackbar, data, hasShownCompleteRegistrationMessage, history, isLoggedIn]);
+
+    return null;
+};
+
+ApplicationMessageNotifier.propTypes = {
+    addSnackbar: PropTypes.func.isRequired,
+};
+
+const mapStateToProps = () => ({});
+const mapDispatchToProps = (dispatch) => ({
+    addSnackbar: (notification) => dispatch(addSnackbar(notification)),
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(ApplicationMessageNotifier);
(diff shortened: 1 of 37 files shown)


Naapperas:
While I was testing this preview locally, when I logged in as a newly created company, a pop-up appeared for me to finish registration for that company, but then it disappeared and I didn't find anywhere for me to be redirected to the "finish company registration" page. Is there any button or link for me to be redirected there after the pop-up? If negative, I suggest adding a button to the "account menu" that only appears to company accounts that haven't finished their registration.

Naapperas:
Don't know if it is the desired behavior, but when testing the first stage of finishing company registration (the "upload logo" one), if I uploaded a photo, pressed the "Upload" button again and then closed the "Select file" pop-up/pressed Escape, the original logo that was uploaded disappears, appearing the "File size must be under 10MB" message. If that is how it is supposed to be, please ignore this comment, but if not, I think maintaining the uploaded logo while no new new is selected could be a nice feature.

Naapperas:
When giving contacts, could we maybe have some sort of validation for the text that is inputted ? I bring this up because of things like this

image

which worries me a bit, as the company could enter gibberish and finish the registration without having given a proper contact.

I don't now how doable this is or if there already is such a validation in the back-end but, if possible and reasonable, please add it.

imnotteixeira:
> While I was testing this preview locally, when I logged in as a newly created company, a pop-up appeared for me to finish registration for that company, but then it disappeared and I didn't find anywhere for me to be redirected to the "finish company registration" page. Is there any button or link for me to be redirected there after the pop-up? If negative, I suggest adding a button to the "account menu" that only appears to company accounts that haven't finished their registration which redirects the user to the appropriate page.

I agree, I will do that, and maybe add a notification dot on the menu so that the users know there's something important there

> Don't know if it is the desired behavior, but when testing the first stage of finishing company registration (the "upload logo" one), if I uploaded a photo, pressed the "Upload" button again and then closed the "Select file" pop-up/pressed Escape, the original logo that was uploaded disappears, appearing the "File size must be under 10MB" message. If that is how it is supposed to be, please ignore this comment, but if not, I think maintaining the uploaded logo while no new new is selected could be a nice feature.

That's a hard thing to do since the cancel action calls the onChange natively. There already is a "Required" validation in place, but probably an empty file object is considered as the input not being empty. I'll try to add a custom validator to check for empty files, but otherwise, there's not much we can do.

> When giving contacts, could we maybe have some sort of validation for the text that is inputted?

That's not on the scope for the MVP, as the API currently does not support it even. In the future, we can have typed contacts (specific formats like email, telephone, website etc), but for now it won't be targetted in this PR

src/components/ApplicationMessages/index.js
```diff
@@ -0,0 +1,60 @@
...
+                            message={message}
+                            actionText="Show me"
+                            actionHandler={() => {
+                                history.push("/company/registration/finish");
+                            }}
+                        />,
+                },
+            });
+        }
+        if (hasShownCompleteRegistrationMessage && !isLoggedIn) setHasShownCompleteRegistrationMessage(false);
```
Naapperas, line 37: What is this line for? I imagine it is useful when the notification has been shown and the user logs out, in which case it is not supposed to show the notification on page "reload", please correct me if I'm wrong.

imnotteixeira: It's to prevent the message from showing multiple times on component remount (not necessarily page reload). Without it, the notification would show every time we go to another page inside this website, since it does not reload (it's a SPA), it simply remounts another subtree of components (which would make this component re-render and show the message again)

This component will also re-render on logout, but the if above would check `isLoggedIn` in order to show the notification. However, if the user would log back in, we want the notification to show again, which would not happen since the `hasShownCompleteRegistrationMessage` is true, so it works kind of like a reset.

Nevertheless, the usage of `hasShownCompleteRegistrationMessage` is important to keep the notification from showing on every re-render (while the user is logged in and without a complete registration)

Naapperas (approved):
LGTM

PR #105, opened 2021-04-23
View offer page

TiagooGomess:
***Updated view in the admin options and blocked companies at the end of this comment***

### Desktop view:

https://user-images.githubusercontent.com/50210213/127509182-2562d958-eacc-41bc-924d-47c0e6f0c099.mp4

### Mobile view:

https://user-images.githubusercontent.com/50210213/127510421-2f64df3b-8e0e-477e-9382-a25b2c97e77e.mp4

### After the last update to the admin options:

https://user-images.githubusercontent.com/50210213/128090584-e7a063ad-2f40-4dae-b414-553ea204b52d.mp4

### If the company is blocked:
Screenshot from 2021-08-03 22-46-36


TiagooGomess:
### Visitor view:
visitor

### Owner company view:
hide-offer

### Admin view:
disable-offer

TiagooGomess:
### When there is an error when hiding / disabling / enabling an offer:

imnotteixeira:
Leaving some generic comments, according to what we discussed in the meeting, so that I don't forget them (without looking at the code yet)

* The title should be in a Grid (taking only 9 slots) and reserve 3 (or maybe 8-4, whatever looks better) for "options", where you would place the "block offer" (for admins) or the Hide offer (admins or owner) buttons. you can use a vertical Divider to make it look better maybe.
* The offer details (location, job type etc) are wrong in terms of grid layout. There should be a single grid container with 3 gird items inside (one for location and duration, other for type and publish date, and the last one for fields and techs). The first 2 should take xs12 and md6, and the last one can only be xs12, since it will occupy a full line always
* Have only one divider between the title and offer details (move the contacts to the offer details section) and another between offer details and offer "content" (which includes offer reqs and description)
* Use `<Avatar>` with default logo (if not present) for company logo
* Use TechOptions and FieldOptions objects for Chip labels
* Tooltips with description of offer details (location, publish date, etc)

* TODO: Check if card width works properly, otherwise we might need to place a maxWidth rule

src/components/HomePage/SearchResultsArea/Offer/OfferContent.js
```diff
@@ -1,22 +1,89 @@
...
+    const classes = useSearchResultsWidgetStyles({ isMobile, isPage, loading });
+
+    const [visibilityState, setVisibilityState] = useState({
+        isVisible: undefined,
+        isDisabled: undefined,
+        isBlocked: undefined,
+    });
+    visibilityState.isDisabled = offer?.isHidden && offer?.hiddenReason === "ADMIN_REQUEST";
+    visibilityState.isVisible = !offer?.isHidden;
+    visibilityState.isBlocked = offer?.isHidden && offer?.hiddenReason === "COMPANY_BLOCKED";
```
imnotteixeira, line 50: This should be computed inside a useEffect, that runs whenever the offer.isHidden or offer.hiddenReason change

src/pages/OfferPage.js
```diff
@@ -0,0 +1,61 @@
...
+            .then((offer) => {
+                setOffer(new Offer(offer));
+                setLoading(false);
+            })
+            .catch(() => {
+                setError("UnexpectedError");
+            });
+        return () => {
+        };
+    }, [id]);
```
imnotteixeira, line 28: Just to make it clearer, this is where the SWR for offers would come in handy. You could create a wrapper hook that called the API through SWR, and you would have a mutate and revalidate functions available for when you want to change the offers (disable, hide etc), which you would pass down through the controller's context (simply return them below.

Then, when the inner components wanted to change the offers (in this case, the AdminReasonModal I think), they can call the functions to mutate locally, and call for a revalidation after the request to hide/disable etc (which will trigger swr to just re-fetch the offer from the API, which should now have the correct values in the fields for isHidden, hiddenReason etc

Be aware that the OfferContent cannot rely on using the context to access these functions, since it is being used on the search widget as well, which does not provide this context. Instead, make the offerContent receive the functions as props, and pass them here, and an alternative function in the search results widget case (which won't be using SWR, but you can still mutate the offers which are stored in the Redux store locally, I believe)

src/components/HomePage/SearchResultsArea/Offer/OfferContent.js
```diff
@@ -28,75 +95,261 @@ const OfferContent = ({ offer, loading }) => {
...
                             Please select an offer to view the details
                         </Typography>
                     </div>
                     :
-                    <React.Fragment>
-                        <div className={classes.offerHeader}>
-                            <Typography variant="h4" gutterBottom>
-                                {offer.title}
-                            </Typography>
+                    <div>
```
imnotteixeira, line 119: This component has gotten too big I think, please subdivide the tree into smaller components (maybe an OfferDetails, and an OfferMessage - actually it should be OfferContent, but this is already OfferContent, unless you come up with a better name... 👀 )

src/components/HomePage/SearchResultsArea/Offer/OfferContent.js
```diff
@@ -1,22 +1,89 @@
...
+            else
+                setVisibilityError(getHumanError(err[0]));
+        }
+        addSnackbar({
+            message: visibilityError ? visibilityError : "Unexpected Error. Please try again later.",
+            key: visibilityError,
+        });
+    };
+
+    const getDisabledOfferMessage = () => (
```
imnotteixeira, line 68: Wrap these functions in `useCallback`, so as to avoid creating a new function every render.

It works like useEffect, you give it the function (what you currently have) as the first arg, and the second is an array of dependences. Whenever the dependencies change, a new function is created. In this case the dependencies should be `sessionData?.isAdmin` and `offer.adminReason`, for example

src/components/HomePage/SearchResultsArea/Offer/OfferContent.spec.js
```diff
@@ -30,32 +56,497 @@ describe("OfferContent", () => {
...
+            useSession.mockImplementation(() => ({ isLoggedIn: true, data: { email: "admin@admin.com", isAdmin: true } }));
+
+            renderWithStoreAndTheme(
+                <OfferContent offer={anotherOffer} />,
+                { theme }
+            );
+
+            expect(screen.queryByText(
+                "This offer is hidden so it won't show up in search results",
+                { exact: false }
```
imnotteixeira, line 272: why do you need the exact:false? What is the exact text?

src/components/HomePage/SearchResultsArea/Offer/OfferContentListItem.js
```diff
@@ -0,0 +1,61 @@
...
+    },
+    list: {
+        margin: 0,
+    },
+    item: {
+        marginTop: theme.spacing(0.5),
+    },
+}));
+
+const OfferContentListItem = ({ title, content, loading }) => {
```
imnotteixeira, line 19:
```suggestion
const OfferContentListItem = ({ title, items, loading }) => {
```

src/components/HomePage/SearchResultsArea/Offer/OfferContentListItem.js
```diff
@@ -0,0 +1,61 @@
...
+                    variant="body1"
+                >
+                    <ul className={classes.list}>
+                        {listItems}
+                    </ul>
+                </Typography>
+            </div>
+        );
+
+    }
```
imnotteixeira, line 51:
```suggestion
    if (loading)
        return (
            <div className={classes.listElement}>
                <Skeleton variant="rect" height="50px" />
            </div>
        );
    else if (items) {
        return (
            <div className={classes.listElement}>
                <Typography variant="h6">
                    {title}
                </Typography>
                <Typography
                    component="span" /* if we don't use component="span",
                                    there is an <ul> element inside a <p>,
                                    which fails the validateDOMNesting*/
                    variant="body1"
                >
                    <ul className={classes.list}>
                        {items.map((item) =>
            <li key={item} className={classes.item}>
                {item}
            </li>
        );}
                    </ul>
                </Typography>
            </div>
        );

    }
```

src/components/HomePage/SearchResultsArea/Offer/OfferDescription.js
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from "prop-types";
+
+import { useEditor, EditorContent } from "@tiptap/react";
+import StarterKit from "@tiptap/starter-kit";
+
+const OfferDescription = ({ content }) => {
+
+    const editor = useEditor({
+        extensions: [
+            StarterKit,
```
imnotteixeira, line 11: Add the Underline extension here as well, to parse underline text. Just install @tiptap/extension-underline and then `import Underline from "@tiptap/extension-underline";`

src/components/HomePage/SearchResultsArea/Offer/OfferVisibilityOptions.js
```diff
@@ -0,0 +1,107 @@
...
+
+const OfferVisibilityOptions = ({
+    loading, sessionData, classes, visibilityState,
+    setVisibilityState, addSnackbar, setShowAdminReasonModal,
+    dealWithPromiseError, isMobile, offer }) => {
+
+    const handleOfferVisibility = () => {
+        if (visibilityState.isVisible) {
+            hideOffer(offer.id).then(() => {
+                offer.isHidden = true;
```
imnotteixeira, line 16: same thing as before, do not mutate offer directly, instead use a function that should be passed from above to mutate the offer object correctly

src/components/HomePage/SearchResultsArea/Offer/OfferVisibilityOptions.js
```diff
@@ -0,0 +1,107 @@
...
+                addSnackbar({
+                    message: "The offer was hidden",
+                    key: "hidden",
+                });
+            }).catch((err) => {
+                dealWithPromiseError(err);
+            });
+        } else  {
+            enableOffer(offer.id).then(() => {
+                offer.isHidden = false;
```
imnotteixeira, line 27: improper object mutation (see above)

src/components/HomePage/SearchResultsArea/Offer/OfferVisibilityOptions.js
```diff
@@ -0,0 +1,107 @@
...
+    };
+
+    const handleEnableDisableOffer = () => {
+        if (!visibilityState.isDisabled) {
+            setShowAdminReasonModal(true);
+        } else {
+            enableOffer(offer.id).then(() => {
+                offer.adminReason = null;
+                offer.hiddenReason = null;
+                offer.isHidden = false;
```
imnotteixeira, line 46: improper object mutation (see above)

src/pages/OfferPage.js
```diff
@@ -0,0 +1,61 @@
...
+    const [loading, setLoading] = useState(true);
+
+    useEffect(() => {
+        getOffer(id)
+            .then((offer) => {
+                setOffer(new Offer(offer));
+                setLoading(false);
+            })
+            .catch(() => {
+                setError("UnexpectedError");
```
imnotteixeira, line 24:
```suggestion
                setError("Unexpected Error. Could not fetch requested offer.");
```

imnotteixeira (changes requested):

src/components/HomePage/SearchResultsArea/Offer/OfferContent.spec.js
TiagooGomess: @imnotteixeira  In this case, if I didn't use the `exact: false`, and for example, the text on the screen was `This offer is hidden so it won't show up in search results.` (with a dot), the test would pass, even though the text `This offer is hidden so it won't show up in search results` was on the screen. If I use the `exact: false`, this problem does not exist. With that said, I think it's a good practice to use the `exact: false` when we are testing if something is not on the screen.

TiagooGomess: https://testing-library.com/docs/queries/about/#textmatch-examples

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsMobile.js
```diff
@@ -128,8 +158,12 @@ const SearchResultsMobile = () => {
                 <OfferViewer
                     open={openPreview}
                     toggleOpenPreview={toggleOpenPreview}
-                    offerContentWrapperClassName={classes.offerBodyContainer}
-                    selectedOffer={selectedOffer}
+                    offerWidgetWrapperClassName={classes.offerBodyContainer}
+                    offer={selectedOfferIdx !== null && offers?.length > 0 ? offers[selectedOfferIdx] : null}
```
imnotteixeira, line 162: Add a condition for `offers?.length > selectedOfferIdx` so that imagining that we have selected the last offer, and redo the search with tighter restrictions (fewer results) it won't crash since offers[selectedIdx] might not exist

imnotteixeira: It should not happen due to the loading state setting the idx to null, but still, it's a nice safeguard for the future (filters being applied might limit the number of offers without triggering the null of the idx)

src/components/HomePage/SearchResultsArea/Offer/AdminReasonModal.js
```diff
@@ -0,0 +1,106 @@
...
+    };
+
+    const { register, handleSubmit, formState: { errors } } = useForm({
+        mode: "onBlur",
+        resolver: yupResolver(AdminReasonModalSchema),
+        reValidateMode: "onChange",
+    });
+
+    const onSubmit = async (data) => {
+        await handleDisableOffer({
```
imnotteixeira, line 39: These functions don't need to know about UI logic such as this setOpen. Don't pass it as argument. Instead call handleClose() after this call (it's an await call so it will close only after it disables)

src/pages/OfferPage.js
```diff
@@ -0,0 +1,168 @@
...
+        if (offerData) {
+            setOffer(new Offer(offerData));
+            setLoading(false);
+        }
+    }, [offerData]);
+
+    const handleDisableOffer = useCallback(async ({
+        offer,
+        adminReason,
+        setOpen,
```
imnotteixeira, line 34: this shouldn't be a param of this function

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,105 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
     // Reset the selected offer on every "loading", so that it does not show up after finished loading
     useEffect(() => {
-        if (offersLoading) setSelectedOffer(null);
+        if (offersLoading) setSelectedOfferIdx(null);
     }, [offersLoading]);
 
+    const handleDisableOffer = useCallback(async ({
+        offer,
+        adminReason,
+        setOpen,
```
imnotteixeira, line 48: this shouldn't be a param of this function

src/services/offerVisibilityService.js
```diff
@@ -0,0 +1,72 @@
+import config from "../config";
```
imnotteixeira, line 1: I think we should just merge this file with the offerSearchService (and call it offerService.js) which would contain offer-related operations.

src/pages/OfferPage.js
```diff
@@ -0,0 +1,168 @@
...
+    enableOffer as enableOfferService,
+} from "../services/offerVisibilityService";
+import useOffer from "../hooks/useOffer";
+
+export const OfferPageControllerContext = React.createContext();
+
+export const OfferPageController = () => {
+    const { id } = useParams();
+    const [offer, setOffer] = useState(null);
+    const [loading, setLoading] = useState(true);
```
imnotteixeira, line 20: You can remove these state variables. The `offer` variable will be the `offerData`, and the loading you can return from the `useOffer` hook (similar to `isValidating`  in the `useSession` hook)

Then you won't need the useEffect below either to sync the variables

imnotteixeira (changes requested):
Final changes to deploy this!

src/pages/OfferPage.js
```diff
@@ -0,0 +1,168 @@
...
+        onError,
+    }) => {
+        await disableOfferService(offer.id, adminReason).then(() => {
+            setOpen(false);
+            mutate({
+                ...offerData,
+                _id: offer.id,
+                hiddenReason: "ADMIN_REQUEST",
+                isHidden: true,
+                adminReason: adminReason,
```
imnotteixeira, line 45:
```suggestion
                adminReason,
```

imnotteixeira (changes requested):

src/components/HomePage/SearchResultsArea/Offer/AdminReasonModal.js
TiagooGomess: @imnotteixeira That way, if the disableOfferService fails, the modal is closed. As it is now, the modal is only closed when that service succeeds.

src/pages/OfferPage.js
TiagooGomess: The first time `useSwr` is called, `isValidating` is false. I'll be using `!data` to set `loading` instead.

Screenshot from 2021-09-02 08-13-36
Screenshot from 2021-09-02 08-13-12

imnotteixeira: I thought isValidating was !data 😅 you can use it that way sure

src/components/HomePage/SearchResultsArea/Offer/AdminReasonModal.js
imnotteixeira: Ok, there are two alternatives then:

- make the function throw on error instead of passing an onError callback, and handle the error when calling this function, then use a try/catch , inside the try you call the function+the UI change (on the admin component which has access to both) and handle thr error on the catch

- pass an onSuccess callback, similar to onError (which in this case the admin modal will tell it to close, but the function doesn't know anything about it) which is only called on success (remember to have something like (onSuccess && onSuccess()) to avoid calling it when it is not given

src/pages/OfferPage.js
imnotteixeira: To be clear, you can export the loading variable from the hook, setting it to !data there (not on the component)

src/components/HomePage/SearchResultsArea/Offer/AdminReasonModal.js
TiagooGomess: @imnotteixeira I implemented the second option, but I don't think it's a good idea to test if onSuccess exists before calling it (`onSuccess && onSuccess()`). `onSuccess` should always exist, and if it doesn't, an error should be thrown, which would not occur if we had that condition. If we followed that logic, we should also test if `onError` exists before calling it.

Screenshot from 2021-09-02 15-45-00
Screenshot from 2021-09-02 15-45-16

imnotteixeira: They both should be optional.

Both callbacks are the way of handleDisableOffer saying: "Do you want to do something on success or on error? I'll do it for you". It doesn't know what it will do **or if it will do anything at all**.

Imagine in the future you stop using a dialog and you don't have to handle the fact it needs to be closed, you won't need to call onSuccess, but it still makes sense from an API point of view (it's the same as throwing the error instead of having onError, you would have an async function, on which you **could** add the .then or .catch to have the exact same behavior, again, optionally)

Besides, the parameters are inherently optional since you have an object which is not required to have all those properties, and since we're not using typescript, there's no way of enforcing them.

Please add the check if callbacks are present before them (on both)

PS: on the last image, when you have `offer: offer, adminReason: data.adminReason, onError: onError, onSuccess: onSuccess`, you can simplify by only having `offer, adminReason, onError, onSuccess` if you destructure the argument `data` by having directly `({ adminReason })` there instead.

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
+        onError,
+    }) => {
+        await hideOfferService(offer.id).then(() => {
+            hideOffer(selectedOfferIdx);
+            addSnackbar({
+                message: "The offer was hidden",
+                key: `${Date.now()}-hidden`,
+            });
+        }).catch((err) => {
+            onError(err);
```
imnotteixeira, line 71:
```suggestion
            if (onError) onError(err);
```

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
+        onError,
+    }) => {
+        await enableOfferService(offer.id).then(() => {
+            companyEnableOffer(selectedOfferIdx);
+            addSnackbar({
+                message: "The offer was enabled",
+                key: `${Date.now()}-enabled`,
+            });
+        }).catch((err) => {
+            onError(err);
```
imnotteixeira, line 87:
```suggestion
            if (onError) onError(err);
```

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
+        addSnackbar,
+        onError,
+    }) => {
+        await enableOfferService(offer.id).then(() => {
+            companyEnableOffer(selectedOfferIdx);
+            addSnackbar({
+                message: "The offer was enabled",
+                key: `${Date.now()}-enabled`,
+            });
+        }).catch((err) => {
```
imnotteixeira, line 86:
```suggestion
        await enableOfferService(offer.id)
            .then(() => {
                companyEnableOffer(selectedOfferIdx);
                addSnackbar({
                    message: "The offer was enabled",
                    key: `${Date.now()}-enabled`,
                });
            })
            .catch((err) => {
```

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
+        addSnackbar,
+        onError,
+    }) => {
+        await hideOfferService(offer.id).then(() => {
+            hideOffer(selectedOfferIdx);
+            addSnackbar({
+                message: "The offer was hidden",
+                key: `${Date.now()}-hidden`,
+            });
+        }).catch((err) => {
```
imnotteixeira, line 70:
```suggestion
        await hideOfferService(offer.id)
            .then(() => {
                hideOffer(selectedOfferIdx);
                addSnackbar({
                    message: "The offer was hidden",
                    key: `${Date.now()}-hidden`,
                });
            })
            .catch((err) => {
```

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
+        addSnackbar,
+        onError,
+    }) => {
+        await enableOfferService(offer.id).then(() => {
+            adminEnableOffer(selectedOfferIdx);
+            addSnackbar({
+                message: "The offer was enabled",
+                key: `${Date.now()}-enabled`,
+            });
+        }).catch((err) => {
```
imnotteixeira, line 102:
```suggestion
        await enableOfferService(offer.id)
            .then(() => {
                adminEnableOffer(selectedOfferIdx);
                addSnackbar({
                    message: "The offer was enabled",
                    key: `${Date.now()}-enabled`,
                });
            })
            .catch((err) => {
```

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
+            onError(err);
+        });
+    }, [companyEnableOffer, selectedOfferIdx]);
+
+    const handleAdminEnableOffer = useCallback(async ({
+        offer,
+        addSnackbar,
+        onError,
+    }) => {
```
imnotteixeira, line 95: I think this fits in a single line now for better readability

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
+        }).catch((err) => {
+            onError(err);
+        });
+    }, [hideOffer, selectedOfferIdx]);
+
+    const handleCompanyEnableOffer = useCallback(async ({
+        offer,
+        addSnackbar,
+        onError,
+    }) => {
```
imnotteixeira, line 79: I think this fits in a single line now for better readability

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsWidget.js
```diff
@@ -11,18 +11,100 @@ import SearchResultsMobile from "./SearchResultsMobile";
...
-        if (offersLoading) setSelectedOffer(null);
+        if (offersLoading) setSelectedOfferIdx(null);
     }, [offersLoading]);
 
+    const handleDisableOffer = useCallback(async ({
+        offer,
+        adminReason,
+        onSuccess,
+        onError,
+    }) => {
```
imnotteixeira, line 50: Check if this fits in a single line now for better readability

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #111, opened 2021-06-30
Added CompanyOffersManagement service

monkin77:
This pr is to create a service for the company offers, in order to fetch offers that belong to a company.


imnotteixeira:
Please use more specific PR titles and descriptions, not just the name of the commit

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -50,15 +51,16 @@ const RowActions = () => (
 
 
 const CompanyOffersManagementWidget = () => {
+    const { data } = useSession();
```
imnotteixeira, line 54:
```suggestion
    const { data, isLoggedIn } = useSession();
```

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -50,15 +51,16 @@ const RowActions = () => (
...
     const [offers, setOffers] = useState(null);
     const [isLoading, setIsLoading] = useState(true);
     const [error, setError] = useState(null);
 
     useEffect(() => {
-        fetchCompanyOffers().then((offers) => {
+        fetchCompanyOffers(data.company._id).then((offers) => {
```
imnotteixeira, line 60: check if `isLoggedIn` before executing

```suggestion
        if(isLoggedIn) fetchCompanyOffers(data.company._id).then((offers) => {
```

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -68,7 +70,7 @@ const CompanyOffersManagementWidget = () => {
             setError(err);
```
imnotteixeira, line 70: where is `err` coming from?

imnotteixeira (changes requested):

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
monkin77: The fetchCompanyOffers method throws an error if the API call fails

src/services/companyOffersService.js
```diff
@@ -1,123 +1,25 @@
...
-    }
-    */
-
+export const fetchCompanyOffers = async (companyID, filters) => {
     await new Promise((resolve) => {
         setTimeout(() => {
             resolve("resolved");
-        }, 1000);
+        }, 500);
+    });
```
imnotteixeira, line 11: Remove unnecessary sleep

imnotteixeira (changes requested):

src/services/companyOffersService.js
```diff
@@ -1,97 +1,12 @@
-import Offer from "../components/HomePage/SearchResultsArea/Offer/Offer";
+import config from "../config";
+const { API_HOSTNAME } = config;
 
 
-// import config from "../config";
-// const { API_HOSTNAME } = config;
+const parseFiltersToURL = (filters) => Object.keys(filters).map((key) => `${key}=${filters[key]}`).join("&");
```
TiagooGomess, line 5: A similar function is defined [here](https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/services/offerService.js#L7). Maybe you can reuse it?

TiagooGomess (changes requested):

src/services/companyOffersService.js
monkin77: I believe it can be reused as well. although for now, the service is not using filters inside the request, it is there for future implementation. Right now, there is the possibility of filtering, but only after fetching the data

monkin77: But I deleted the duplicated method already to avoid confusion in the future

src/services/offerService.js
```diff
@@ -4,7 +4,7 @@ import Offer from "../components/HomePage/SearchResultsArea/Offer/Offer";
 import config from "../config";
 const { API_HOSTNAME } = config;
 
-const parseSearchFiltersToURL = (filters) => Object.keys(filters)
+export const parseSearchFiltersToURL = (filters) => Object.keys(filters)
```
TiagooGomess, line 7: I suggest moving this function to the `/utils` folder and renaming it to `parseFiltersToURL` since it can be reused in other places.

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -51,16 +52,17 @@ const RowActions = () => (
...
 const CompanyOffersManagementWidget = () => {
+    const { data, isLoggedIn } = useSession();
     const [offers, setOffers] = useState(null);
     const [isLoading, setIsLoading] = useState(true);
     const [error, setError] = useState(null);
 
     useEffect(() => {
-        fetchCompanyOffers().then((offers) => {
+        if (isLoggedIn) fetchCompanyOffers(data.company._id).then((offers) => {
             /* TODO: SHOULD NOT RUN WHEN COMPONENT IS UNMOUNTED: CANCEL PROMISE */
```
TiagooGomess, line 62: When will this be done? Maybe we can open an issue to reference this.

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -6,12 +6,15 @@ import { renderWithStore } from "../../../../test-utils";
...
+jest.mock("../../../../services/companyOffersService");
 
 describe("App", () => {
     const MOCK_OFFERS = [
         {
-            id: "random uuid4",
+            _id: "random uuid4",
             title: "Guy in the background",
             ownerName: "Reddit",
             ownerLogo: "logo.com",
```
TiagooGomess, line 20: The mock offers should have an `owner` property referring to the id of the company.

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -21,7 +24,7 @@ describe("App", () => {
             description: "kek",
         },
         {
-            id: "random uuid5",
+            _id: "random uuid5",
             title: "Guy in the background",
             ownerName: "Reddit",
             ownerLogo: "logo.com",
```
TiagooGomess, line 30: The mock offers should have an `owner` property referring to the id of the company.

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -32,7 +35,15 @@ describe("App", () => {
         },
     ];
 
+    beforeEach(() => {
+        useSession.mockReturnValue({ data: { company: "company" }, isLoggedIn: true });
```
TiagooGomess, line 39: The company id is missing. Do this instead:
```js
useSession.mockImplementation(() => ({
    isLoggedIn: true, data: { company: { name: "company1", _id: "company_id" } },
}));
```
Don't forget that the company id (in this case "company_id") should be equal to the `owner` property of the mocked offers.

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
```diff
@@ -8,6 +8,7 @@ import { generateTableCellFromField } from "../../../../utils/Table/utils";
 import { columns } from "./CompanyOffersManagementSchema";
 import PropTypes from "prop-types";
 import { format, parseISO } from "date-fns";
+import useSession from "../../../../hooks/useSession";
 
 const CompanyOffersNonFullfilledRequest = ({ isLoading, error }) => {
```
TiagooGomess, line 13:
```suggestion
const CompanyOffersNonFulfilledRequest = ({ isLoading, error }) => {
```

TiagooGomess: Rename the usages of this component accordingly.

TiagooGomess (changes requested):

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -75,13 +91,15 @@ describe("App", () => {
...
         });
     });
 
     test("Error fetching offers", async () => {
-        companyOffersService.fetchCompanyOffers = jest.fn(() => new Promise(() => {
-            throw "Error fetching offers";
-        }));
+        companyOffersService.fetchCompanyOffers.mockImplementationOnce(() => new Promise((resolve, reject) => setTimeout(() => {
+            reject("Error fetching offers");
+        }, 1000)));
```
imnotteixeira, line 102: Why not reject immediately? These types of mocks (with the set timeout) make the test run much longer, and unnecessarily in my opinion

imnotteixeira (changes requested):

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.js
imnotteixeira: That is done in #96, kind of my bad for letting that rest for half a year :grimacing: 

In my defense, it didn't have that much priority ahaha

monkin77:
@TiagooGomess Do you know why the tests are failing? They don't seem to be related with this branch and when I run locally there's no error

imnotteixeira:
> @TiagooGomess Do you know why the tests are failing? They don't seem to be related with this branch and when I run locally there's no error

You still need to remove the setTimeout calls in the resolve/reject in your tests, as well as the tests' timeouts `it("should...", ()=>{}, TIMEOUT HERE; REMOVE)`

monkin77:
> > @TiagooGomess Do you know why the tests are failing? They don't seem to be related with this branch and when I run locally there's no error
> 
> You still need to remove the setTimeout calls in the resolve/reject in your tests, as well as the tests' timeouts `it("should...", ()=>{}, TIMEOUT HERE; REMOVE)`
I've removed the setTimeout calls where it was waiting. but what are the test timeouts you're referring to? is it the ones inside the waitFor method? If so, currently it is stopping execution if the conditions are met before the time

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -32,7 +37,15 @@ describe("App", () => {
...
     ];
 
+    beforeEach(() => {
+        useSession.mockReturnValue({ data: { company: { name: "company1", _id: "company_id" } }, isLoggedIn: true });
+    });
+
     test("Renders Loading", () => {
+        companyOffersService.fetchCompanyOffers.mockImplementationOnce(() => new Promise((resolve) => setTimeout(() => {
+            resolve();
+        }, 1000)));
```
imnotteixeira, line 47: This test is dependant on the fact that the render takes less than 1000ms, that may not happen due to concurrent test execution making it slower, or some other reason. Instead, don't call the resolve function (remove the setTimeout as well. This way you return a Promise that will never resolve, and it will be in loading state forever

imnotteixeira (changes requested):

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
monkin77: Yeah I was thinking about refactoring this test as well :smile:.. Done

src/components/Company/Offers/Manage/CompanyOffersManagementWidget.spec.js
```diff
@@ -60,12 +73,15 @@ describe("App", () => {
...
+            timeout: 1000,
         });
     });
 
     test("Loads Empty Offers", async () => {
-        companyOffersService.fetchCompanyOffers = jest.fn(() =>  Promise.resolve([]));
-
+        companyOffersService.fetchCompanyOffers.mockImplementationOnce(() =>  new Promise((resolve) =>
+            resolve([])
+        ));
```
imnotteixeira, line 84: You can return a "resolved" promise instantly like this

```suggestion
        companyOffersService.fetchCompanyOffers.mockImplementationOnce(() =>  Promise.resolve([])));
```

imnotteixeira (approved):

TiagooGomess (approved):

PR #123, opened 2021-09-03
Fixed finish registration badge showing for admin

### LINKED ISSUE #112: Finish registration badge is showing for admin
This account is the default admin, it shouldn't have to complete any registration.

image

TiagooGomess:
Closes #112

### DIFF
--- a/src/components/Navbar/UserMenu.js
+++ b/src/components/Navbar/UserMenu.js
@@ -86,6 +86,37 @@ AdminMenuOptions.propTypes = {
     isMobile: PropTypes.bool.isRequired,
 };
 
+const CompanyMenuOptions = ({ isMobile, sessionData }) => {
+
+    const classes = useStyles({ isMobile });
+
+    if (sessionData.company.hasFinishedRegistration)
+        return (
+            <MenuItem button component={Link} to="/company/offers/manage" disableTouchRipple>My Offers</MenuItem>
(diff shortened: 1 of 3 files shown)


src/components/Navbar/UserMenu.js
```diff
@@ -86,6 +86,37 @@ AdminMenuOptions.propTypes = {
...
+                    color="primary"
+                    anchorOrigin={{ horizontal: "left", vertical: "top" }}
+                >
+                    <Box pl={2}>
+                        Finish Registration
+                    </Box>
+                </Badge>
+            </MenuItem>
+        );
+    }
```
imnotteixeira, line 117: This is not very future proof (if we want to add a new option that shows regardless of that condition you would have to add it to both branches (if/else)

You can do it like this instead:
```
return (
<>
{sessionData.company.hasFinishedRegistration ? 
    <MyOffers stuff>
    : <FinishRegistration stuff>
}
// stuff that goes here shows every time
</>
)
```

imnotteixeira: Also use `sessionData?.company?.hasFinishedRegistration` since if the logged in user is an admin, it would crash

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #127, opened 2021-09-09
Lost state fix

### LINKED ISSUE #115: The state is lost when switching from desktop to mobile view and vice-versa
The session was being lost in some pages, so we needed to set revalidateOnMount to true [here](https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/components/Navbar/index.js#L30), in #105.

Because of this change, the state is being lost in some pages when switching from desktop to mobile view or vice-versa, namely in the finish registration page.

What is happening is:
1. Unmount of PageLayout
2. Show loading
3. Mount PageLayout again

A possible solution for this is to do not let PageLayout re-render, even if the session changes. In theory, we can try to move the Controller Pattern to the Route level, but we need to see if that would work.

TiagooGomess:
Closes #115

### DIFF
--- a/src/components/Layout/PageLayout.js
+++ b/src/components/Layout/PageLayout.js
@@ -4,8 +4,7 @@ import { useMobile } from "../../utils/media-queries";
 import DesktopLayout from "./DesktopLayout";
 import MobileLayout from "./MobileLayout";
 import BaseLayout from "./BaseLayout";
-
-const DefaultContext = React.createContext();
+import { DefaultContext } from "../../hooks/useComponentController";
 
 export const LayoutType = Object.freeze({
     NONE: "NONE",
@@ -42,8 +41,7 @@ const PageLayout = ({
         document.title = pageTitle ? `${pageTitle} - NIJobs` : "NIJobs";
     }, [pageTitle]);
 
-    const usedContext = context ? context : DefaultContext;
-    const contextValues = useContext(usedContext || DefaultContext);
+    const contextValues = useContext(context || DefaultContext);
 
     const isMobileSize = useMobile();
     const shouldUseMobileLayout = (shouldShowMobile === undefined) ? isMobileSize : shouldShowMobile({ ...contextValues, isMobileSize });
--- a/src/components/Layout/PageLayout.js
+++ b/src/components/Layout/PageLayout.js
@@ -5,7 +5,7 @@ import DesktopLayout from "./DesktopLayout";
 import MobileLayout from "./MobileLayout";
 import BaseLayout from "./BaseLayout";
 
-export const DefaultContext = React.createContext();
+const DefaultContext = React.createContext();
 
 export const LayoutType = Object.freeze({
     NONE: "NONE",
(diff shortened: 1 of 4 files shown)


src/components/Layout/PageLayout.js
```diff
@@ -1,11 +1,12 @@
-import React, { useEffect } from "react";
+import React, { useContext, useEffect } from "react";
 import PropTypes from "prop-types";
 import { useMobile } from "../../utils/media-queries";
-import useComponentController from "../../hooks/useComponentController";
 import DesktopLayout from "./DesktopLayout";
 import MobileLayout from "./MobileLayout";
 import BaseLayout from "./BaseLayout";
 
+const DefaultContext = React.createContext();
```
imnotteixeira, line 8: import the default context from the useComponentController file (https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/hooks/useComponentController.js) It's the same for now, but if we ever need to change it, it's better to use the same

src/components/Layout/PageLayout.js
```diff
@@ -31,32 +32,31 @@ const PageLayout = ({
...
 }) => {
 
-    const { ContextProvider, contextProviderProps, ...contextValues } = useComponentController(controller, controllerProps, context);
-
     useEffect(() => {
         document.title = pageTitle ? `${pageTitle} - NIJobs` : "NIJobs";
     }, [pageTitle]);
 
+    const usedContext = context ? context : DefaultContext;
+    const contextValues = useContext(usedContext || DefaultContext);
```
imnotteixeira, line 46:
```suggestion
    const contextValues = useContext(context || DefaultContext);
```

src/utils/index.js
```diff
@@ -127,35 +128,72 @@ const mapDispatchToProps = (dispatch) => ({
 });
 const RedirectInfoProvider = connect(mapStateToProps, mapDispatchToProps)(BaseRedirectInfoProvider);
 
-export const Route = ({ children, ...props }) => (
+/* istanbul ignore next */
```
imnotteixeira, line 131: why is this needed?

src/utils/index.js
```diff
@@ -191,10 +230,57 @@ export const ProtectedRoute = ({
...
+    controllerProps,
+    // eslint-disable-next-line react/prop-types
+    context,
+    ...routeProps
+}) => (
+    <Route
+        {...routeProps}
+    >
+        <ProtectedRouteController
+            isProtected
```
imnotteixeira, line 270:
```suggestion
```

imnotteixeira (changes requested):

src/utils/index.spec.js
```diff
@@ -47,24 +52,214 @@ describe("utils", () => {
...
                 <Wrap on={false} Wrapper={WrapperComponent}>
                     <p>Hello</p>
                 </Wrap>
             );
 
-            expect(wrapper.queryByTestId("test")).not.toBeInTheDocument();
-            expect(wrapper.getByText("Hello")).toBeInTheDocument();
+            expect(screen.queryByTestId("test")).not.toBeInTheDocument();
+            expect(screen.getByText("Hello")).toBeInTheDocument();
+        });
```
imnotteixeira, line 73: nice 👍

imnotteixeira (approved):
Nice tests, good job 🚀

PR #133, opened 2021-09-21
Add service metrics 

### LINKED ISSUE #55: Add service metrics
Metrics are the most important part of an application, aside from the core functionality. It's the a good and easy way for developers to learn how customers are using their app, and isolate pain points.

As a first effort, we would like to have information on the following events:

- Offer creation
  - Time to create (ms) (from the moment the user clicks on the button to the success response)
  - Errors (every time there is an error, generate an event, so we can count the events later) 
- Offer search
  - Time to result (ms) (from the moment the user clicks on the button to the success response)
  - Errors (every time there is an error, generate an event, so we can count the events later)

For every page, generate an event every time the page is loaded so, that way, we have an idea about the load/usage of the app.

In the beginning, I did a very small PoC with Google analytics, which proved capable of handling the requested functionality. However, if you think there is a better alternative, please justify and we can discuss it.

BrunoRosendo:
closes #55

### DIFF
--- a/.env
+++ b/.env
@@ -6,4 +6,5 @@ REACT_APP_API_HOSTNAME="https://localhost:8087"
 # Uncomment to disable devtools
 # REACT_APP_ALLOW_DEV_TOOLS=false
 
-ANALYTICS_ID="UA-207901526-1"
+# Google Analytics' Universal ID
+ANALYTICS_ID="UA-208192118-1"
--- a/.env
+++ b/.env
@@ -5,3 +5,5 @@ REACT_APP_API_HOSTNAME="https://localhost:8087"
 
 # Uncomment to disable devtools
 # REACT_APP_ALLOW_DEV_TOOLS=false
+
+ANALYTICS_ID="UA-207901526-1"
--- a/src/services/accountService.js
+++ b/src/services/accountService.js
@@ -1,8 +1,9 @@
 import config from "../config";
-import { measureTime, createEvent, TIMED_ACTIONS, EVENT_TYPES } from "../utils/AnalyticsUtils";
+import { measureTime, createEvent, TIMED_ACTIONS, EVENT_TYPES } from "../utils/analytics";
+import ErrorTypes from "../utils/ErrorTypes";
 const { API_HOSTNAME } = config;
 
-const COMPLETE_REGISTRATION = "Complete Registration";
+const COMPLETE_REGISTRATION_METRIC_ID = "registration/complete";
 
 
 export const completeRegistration = measureTime(TIMED_ACTIONS.COMPLETE_REGISTRATION, async ({ logo, bio, contacts }) => {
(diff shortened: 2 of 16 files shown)


.env
```diff
@@ -4,4 +4,7 @@ HOST_PORT=443
 REACT_APP_API_HOSTNAME="https://localhost:8087"
 
 # Uncomment to disable devtools
-# REACT_APP_ALLOW_DEV_TOOLS=false
\ No newline at end of file
+# REACT_APP_ALLOW_DEV_TOOLS=false
+
+# Google Analytics' Universal ID
+ANALYTICS_ID="UA-208192118-1"
```
imnotteixeira, line 10: Is this public? :grimacing:

imnotteixeira: In case it's not, see if you can generate a different one and don't push it to github

src/services/accountService.js
```diff
@@ -21,14 +22,25 @@ export const completeRegistration = async ({ logo, bio, contacts }) => {
         const json = await res.json();
 
         if (!res.ok) {
+
+            createEvent(EVENT_TYPES.ERROR(
+                "Complete Registration",
```
imnotteixeira, line 27: Make this a constant at the top of the file, so that all events here are consistent.

(This goes for all service metrics in other files as well)

src/services/accountService.js
```diff
@@ -21,14 +22,25 @@ export const completeRegistration = async ({ logo, bio, contacts }) => {
...
+                "BAD_RESPONSE",
+                res.status
+            ));
+
```
imnotteixeira, line 31: We also want to have "success" metrics (after this if). With this we would know how many registrations we gain over time and check possible trends

imnotteixeira: Again, success metrics apply to other services as well

src/services/offerService.js
```diff
@@ -33,27 +35,35 @@ export const searchOffers = (filters) => async (dispatch) => {
...
+            ));
             return;
         }
         const offers = await res.json();
         dispatch(setSearchOffers(offers.map((offerData) => new Offer(offerData))));
-
         dispatch(setLoadingOffers(false));
-        // TODO count metrics
+
+        recordTime(TIMED_ACTIONS.OFFER_SEARCH, t0, performance.now());
```
imnotteixeira, line 50: This is perhaps the most important service on which to collect metrics. It might be interesting to add metrics on which filters are being used, for example to get a feeling of what students look for

src/utils/AnalyticsUtils.js
```diff
@@ -0,0 +1,123 @@
+import ReactGa from "react-ga";
+
+/**
+ * Initializes Google Analytics.
+ */
+export const initAnalytics = () => {
+    ReactGa.initialize(process.env.ANALYTICS_ID, {
+        debug: true, // enables useful info on console
```
imnotteixeira, line 8: Don't forget to remove this before merge

imnotteixeira (changes requested):
Good start! Can't wait to see what this will look like when live

.env
BrunoRosendo: Yes, we can create a new one once we release NiJobs :p

imnotteixeira (approved):

PR #136, opened 2021-09-24
Fixed x overflow in mobile view

### LINKED ISSUE #135: Mobile view x overflow
When we scroll horizontally in the mobile view, we can see the overflow of the page:

https://user-images.githubusercontent.com/50210213/134698540-26305757-754d-4e63-82cb-34032f21f10d.mp4

We need to remove the x overflow.

TiagooGomess:
Closes #135

### DIFF
--- a/src/components/HomePage/ProductPlacementArea/productDescriptionStyles.js
+++ b/src/components/HomePage/ProductPlacementArea/productDescriptionStyles.js
@@ -1,7 +1,7 @@
 /* istanbul ignore file */
 import { makeStyles } from "@material-ui/core/styles";
 
-export default (isMobile) => makeStyles(() => ({
+export default (isMobile) => makeStyles((theme) => ({
     productDescription: {
         display: "flex",
         justifyContent: "center",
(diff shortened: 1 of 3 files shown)


TiagooGomess:
I don't know if this is the better approach to fix this problem, I need some feedback.

imnotteixeira:
This is a hack, please find the element that is triggering the overflow and fix that instead

imnotteixeira:
> This is a hack, please find the element that is triggering the overflow and fix that instead

From what I could gather, it's the Grid from ProductDescription

imnotteixeira:
The fix is to remove the `spacing={10}` from the outer Grid container, but that will make the items be closer together, so you need to add padding to them specifically, to maintain the visual aspect

src/components/HomePage/ProductPlacementArea/productDescriptionStyles.js
```diff
@@ -23,5 +23,9 @@ export default makeStyles(() => ({
     companyName: {
         fontWeight: "bold",
     },
-}))
-;
+    productDescriptionGrid: {
+        "& > div": {
+            padding: isMobile ? "40px 0" : "40px 40px",
```
imnotteixeira, line 28: Use theme.spacing instead of hardcoded 40px

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #150, opened 2022-01-04
Added a divider before offer-description

### LINKED ISSUE #146: Add Divider Before Offer Description
Add a divider before the offer description to the offer page, similar to the ones already present.

[Here](https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/components/HomePage/SearchResultsArea/Offer/OfferContent.js#L13) is an example of a divider on this page.

CarlosMealha:
closes #146 

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ typings/
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# .vscode directory
+.vscode/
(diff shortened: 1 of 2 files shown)


imnotteixeira (commented):
Code seems fine, can you post a screenshot of the change to verify it works (easier than having to setup everything to have offers in a custom backend of mine)

Also, "uncommit" package-lock. Those changes are unrelated to this, and it should never be modified except when directly modifying dependencies.

imnotteixeira (changes requested):
Code seems fine, can you post a screenshot of the change to verify it works (easier than having to setup everything to have offers in a custom backend of mine)

Also, "uncommit" package-lock. Those changes are unrelated to this, and it should never be modified except when directly modifying dependencies.

imnotteixeira (approved):

PR #159, opened 2022-01-15
feat: style create offer page

### LINKED ISSUE #151: Style create offer page
Adjust the padding/margins:

Justify the Job Start Date and the Vacancies:

image

limwa:
## Description

Closes #151.
This pull request styles the Create Offer page. 

White-space was added in between every grid component. 
Some additional white-space was also added after the Advanced Settings dropdown, the Contacts input and the Requirements input. After the Vacancies input, I've also added some white-space so that it remains consistent on taller displays and in mobile view.

Finally, I've also added some white-space bellow the submit button because, to me, it felt like there was not enough of it. If you don't agree with this change, let me know and I'll happily remove it.

## Notes

There are still some things I am not happy with and that I wasn't able to fix, such as:

- There is too much space above "Technologies"
image

- When the screen is small, while being bigger than a mobile display, the card shadow looks weird (this could be just me, so let me know your thoughts on this)
image

### DIFF
--- a/src/components/Offers/New/CreateOfferForm.js
+++ b/src/components/Offers/New/CreateOfferForm.js
@@ -76,7 +76,7 @@ export const CreateOfferController = () => {
             contacts: [{ value: "" }],
             isPaid: "none",
             vacancies: "",
-            jobType: "",
+            jobType: "SUMMER INTERNSHIP",
             fields: [],
             technologies: [],
             location: null,
@@ -234,7 +234,7 @@ const CreateOfferForm = () => {
                                 onSubmit={submit}
                                 aria-label="Create Offer Form"
                             >
-                                <Grid container>
+                                <Grid container spacing={4}>
                                     <Grid item xs={12} lg={isAdmin ? 6 : 12}>
                                         <TitleComponent
                                             disabled={disabled}
@@ -262,13 +262,20 @@ const CreateOfferForm = () => {
                                             />
                                         </FormControl>
                                     </Grid>
-                                    <Grid item xs={12} lg={6}>
+                                    <Grid item xs={12} lg={6} className={classes.jobTypeGrid}>
                                         <FormControl fullWidth margin="dense">
                                             <JobTypeComponent
                                                 disabled={disabled}
                                                 errors={errors}
                                                 requestErrors={requestErrors}
                                                 control={control}
+                                                textFieldProps={{
+                                                    SelectProps: {
+                                                        classes: {
+                                                            select: classes.menuSelect,
+                                                        },
+                                                    },
+                                                }}
                                             />
                                         </FormControl>
                                     </Grid>
@@ -283,6 +290,12 @@ const CreateOfferForm = () => {
                                                     onBlur={onBlur}
                                                     error={errors.fields || requestErrors.fields}
                                                     disabled={disabled}
+                                                    chipWrapperProps={{
+                                                        className: classes.autocompleteChipWrapper,
+                                                    }}
+                                                    textFieldProps={{
+                                                        margin: "none",
+                                                    }}
                                                     {...FieldsSelectorProps}
                                                 />
                                             )}
@@ -300,13 +313,19 @@ const CreateOfferForm = () => {
                                                     onBlur={onBlur}
                                                     error={errors.technologies || requestErrors.technologies}
                                                     disabled={disabled}
+                                                    chipWrapperProps={{
+                                                        className: classes.autocompleteChipWrapper,
+                                                    }}
+                                                    textFieldProps={{
+                                                        margin: "none",
+                                                    }}
                                                     {...TechSelectorProps}
                                                 />)}
                                             control={control}
                                         />
                                     </Grid>
                                     <Grid item xs={12} lg={6}>
-                                        <FormControl>
+                                        <FormControl fullWidth>
                                             <JobStartDateComponent
                                                 disabled={disabled}
                                                 errors={errors}
@@ -315,7 +334,7 @@ const CreateOfferForm = () => {
                                             />
                                         </FormControl>
                                     </Grid>
-                                    <Grid item xs={12} lg={6}>
+                                    <Grid item xs={12} lg={6} className={classes.jobDurationGrid}>
                                         <JobDurationComponent
                                             disabled={disabled}
                                             errors={errors}
@@ -323,8 +342,8 @@ const CreateOfferForm = () => {
                                             control={control}
                                         />
                                     </Grid>
-                                    <Grid item xs={12} lg={6}>
-                                        <FormControl>
+                                    <Grid item xs={12} lg={6} className={classes.vacanciesGrid}>
+                                        <FormControl fullWidth>
                                             <VacanciesComponent
                                                 disabled={disabled}
                                                 errors={errors}
@@ -339,6 +358,13 @@ const CreateOfferForm = () => {
                                             errors={errors}
                                             requestErrors={requestErrors}
                                             control={control}
+                                            textFieldProps={{
+                                                SelectProps: {
+                                                    classes: {
+                                                        select: classes.menuSelect,
+                                                    },
+                                                },
+                                            }}
                                         />
                                     </Grid>
                                     <Grid item xs={12} lg={12}>
@@ -346,7 +372,6 @@ const CreateOfferForm = () => {
                                             onClick={() => setAdvancedOpen(!isAdvancedOpen)}
                                             size="small"
                                             margin="dense"
-                                            style={{ marginTop: 10 }}
                                             endIcon={
                                                 isAdvancedOpenOrErrors()
                                                     ? <KeyboardArrowUp />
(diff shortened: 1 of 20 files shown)


limwa:
I'll elaborate on what I've said about there being too much space above the "Technologies" input:

The reason it is hard to fix is because the input, by default, has a pretty big margin on top. On top of that, the Grid component doesn't support row spacing and column spacing in MUI v4. 
I have tried many solutions such as using negative margins (which I hate) and it doesn't work (on the mobile screen it looks good, but on the normal screen it is misaligned). I have also tried wrapping the grid that contains those elements in another grid and it defeats the purpose since we'd have to add even more spacing. 

Do you have any ideas on how we could fix this?

limwa:
[This line](https://github.com/NIAEFEUP/nijobs-fe/blob/d6d0d2fd0f08c6600857259112cc6dc11ea27067/src/components/utils/form/MultiOptionAutocomplete.js#L52) is what is causing these fields to have default margins. We could change it do "dense", however, I am worried that there would be problems in other pages...

DoStini:
Regarding the second question, you can customize the input with inputProps. You can't probably change margin, but you can create a textFieldProps, that follows the same logic as inputProps. You can apply those props as the last parameter of TextField with `{...textFieldProps}` and modify the margin in the create offer page

src/components/Offers/New/createOfferStyles.js
```diff
@@ -37,8 +36,35 @@ export default (isMobile) => makeStyles((theme) => ({
...
+        marginBottom: theme.spacing(4),
+    },
+    fieldsGrid: {
+        marginBottom: theme.spacing(1),
+    },
+    technologiesGrid: {
+        marginBottom: theme.spacing(1),
+    },
+    jobDurationGrid: {
+        paddingLeft: `${theme.spacing(2)}px !important`,
```
DoStini, line 66: Why is the important necessary here?

src/components/utils/form/MultiOptionAutocomplete.js
```diff
@@ -20,6 +20,7 @@ const MultiOptionAutocomplete = forwardRef(
         className,
         disabled,
         inputProps,
+        margin,
```
DoStini, line 23: I think a more flexible solution is to replace this with a `textFieldProps`, and in the last argument of `TextField`, use `...textFieldProps`

DoStini (changes requested):

DoStini (changes requested):
I think it looks more uniform if we remove this [margin](https://github.com/NIAEFEUP/nijobs-fe/blob/59308ba9ca47ee54dd4d3eeaf5e736e6c52a90ec/src/components/Offers/New/form-components/TitleComponent.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L27)

limwa:
Things on the todo list:
- [ ] increase grid spacing to 3 or 4
- [ ] make the start date, duration, vacancies and compensation have the same width and centered (put them on the same grid)
- [ ] make the spacing between Advanced Settings and Contacts a lot smaller (24px or something like that)
- [ ] try to make job duration slider similar to the other ones

limwa:
The changes we've talked about on Slack have been implemented.

Also, I have added the space required below "Fields" and "Technologies" so that the first line of chips doesn't shift the content of the page. To be honest, I feel like there's too much space there and shifting the content would be preferable, but let me know what you think.

Regarding the request to decrease the space below "Advanced Settings", I think it is probably best to leave it as it is. The Collapse element has a grid with spacing 4 in order to keep the publish and publish end dates apart with the same distance as the content grid, which also ends up affecting the "Hide Offer" checkbox. If there was a smaller space below Advanced Settings, the spacing between the dates and hide offer would be much bigger than the space between hide offer and the contacts field (and in my opinion, it looks quite weird and feels like it doesn't belong there).

If you have any questions or suggestions, let me know.

limwa:
I almost forgot one thing: regarding the * to show which fields are mandatory, I'm not quite sure what is the best way to do it. Do we just place an asterisk on the placeholder of the inputs?

src/components/Offers/New/CreateOfferForm.js
```diff
@@ -76,7 +76,7 @@ export const CreateOfferController = () => {
             contacts: [{ value: "" }],
             isPaid: "none",
             vacancies: "",
-            jobType: "",
+            jobType: "SUMMER INTERNSHIP",
```
DoStini, line 79: Was this for testing?

src/components/Offers/New/CreateOfferForm.js
```diff
@@ -262,13 +262,20 @@ const CreateOfferForm = () => {
...
                                                 disabled={disabled}
                                                 errors={errors}
                                                 requestErrors={requestErrors}
                                                 control={control}
+                                                textFieldProps={{
+                                                    SelectProps: {
+                                                        classes: {
+                                                            select: classes.menuSelect,
+                                                        },
+                                                    },
```
DoStini, line 277: unwrap this object to a variable since its being repeated

src/components/Offers/New/CreateOfferForm.js
```diff
@@ -339,14 +358,20 @@ const CreateOfferForm = () => {
                                             errors={errors}
                                             requestErrors={requestErrors}
                                             control={control}
+                                            textFieldProps={{
+                                                SelectProps: {
+                                                    classes: {
+                                                        select: classes.menuSelect,
+                                                    },
+                                                },
```
DoStini, line 366: use variable

DoStini:
I'm thinking now, maybe the last value should be for example 12+ months

DoStini:
Regarding the required fields, did you have something in mind? @TiagooGomess

TiagooGomess:
> I almost forgot one thing: regarding the * to show which fields are mandatory, I'm not quite sure what is the best way to do it. Do we just place an asterisk on the placeholder of the inputs?

@limwa Yes, something similar to this:

image

limwa:
> I'm thinking now, maybe the last value should be for example 12+ months

Sorry, I didn't understand.
What do you mean by that?

TiagooGomess:
> > I'm thinking now, maybe the last value should be for example 12+ months
> 
> Sorry, I didn't understand. What do you mean by that?

I think @DoStini was referring to the job duration upper limit. It should be 12+ months instead of 12.
image

DoStini (changes requested):

limwa:
@DoStini @TiagooGomess I have implemented all your requested changes!
Furthermore, I have readded the Any job option because the job offer search relies on it. I also think that, in the future, we should consider adding an X button to the Job Type field in the job offer search instead of having Any as a job option.

TiagooGomess (approved):

DoStini:
Regarding what I said before about 12+, I think we need to do it as well in the offer page:

image

limwa:
> Regarding what I said before about 12+, I think we need to do it as well in the offer page:
> 
> image

I agree. Isn't it out of this PR's scope, though?
Either way, extracting [these lines](https://github.com/NIAEFEUP/nijobs-fe/blob/d406280b557ad6a93512cacc1a7b4949222655fa/src/components/Offers/New/form-components/JobDurationComponent.js#L43-L45) to a separate function would help standardize the way we display a job offer's duration in NIJobs.

DoStini (approved):

TiagooGomess (approved):
Nice job! Rebase with `develop` and it's ready. :rocket:

TiagooGomess:
The Branch Test Coverage needs to be >= 80%. RIP

src/components/Offers/New/form-components/JobTypeComponent.js
```diff
@@ -15,25 +15,29 @@ const JobTypeComponent = ({ disabled, errors, requestErrors, control }) => (
...
-                        value={value}
-                    >
-                        {label}
-                    </MenuItem>
-                ))}
+                {JobOptions
+                    .filter((option) => option.value !== null)
+                    .map(({ value, label }) => (
+                        <MenuItem
+                            data-testid="job-type-item"
```
DoStini, line 33: Do we really need this data testid?

src/components/Offers/New/CreateOfferForm.spec.js
```diff
@@ -204,13 +203,37 @@ describe("Create Offer Form", () => {
...
+            await act(async () => {
+                const button = await screen.findByLabelText("Job Type *");
+                fireEvent.mouseDown(button);
+            });
+
+            const elements = await screen.findAllByTestId("job-type-item");
+            elements.forEach((element) => {
+                expect(element).toHaveAttribute("data-value");
+                expect(element).toBeVisible();
+            });
```
DoStini, line 234: Instead of test id, maybe something like `JobOptions.forEach((key, val) => screen.findByText(val)` would work

src/components/Offers/New/form-components/JobTypeComponent.js
limwa: As we discussed on slack, that is needed because the objective is to test if we don't have null data-values.

src/components/Offers/New/CreateOfferForm.spec.js
limwa: Please check out [this comment](https://github.com/NIAEFEUP/nijobs-fe/pull/159#discussion_r825454975)

DoStini (approved):

PR #168, opened 2022-01-31
Bump node version

### LINKED ISSUE #167: Bump node version to 14
Update the Dockerfile so that our project uses node 14.

TiagooGomess:
Closes #167


DoStini (approved):
image

The problem is still not fixed, so it's not about node version

PR #172, opened 2022-02-11
Improve Search Functionality

### LINKED ISSUE #79: Improve search functionality
After #78, basic search functionality will be added. However, there are no pagination/smart loading of results in place, which, in the event of being too many, will never exceed the hard limit of 100 imposed by the backend at the moment, forcing the frontend to request future values manually (simulating different pages).

This is why we should implement some sort of smart loading of results (infinite scroll) where we load them in small batches as the user scrolls the list of results. To help with performance on the browser, we should also add virtualization to that list so that even though we have already fetched 200 results, the DOM only contains the visible ones plus some offset (let's say 20 total), thus preventing the page to have to deal with a lot of results which are not even being rendered and prevent lag in the UI.

TiagooGomess:
https://user-images.githubusercontent.com/50210213/153931139-97eec644-55bc-40b2-bc1a-4e2b9148ee4a.mp4

Closes #79


TiagooGomess:
The only thing missing is the implementation of new tests.

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchResultsUtils.js
```diff
@@ -0,0 +1,4 @@
+export const SearchResultsConstants = {
+    initialLimit: 5,
+    fetchNewOffersLimit: 5,
```
imnotteixeira, line 3: Make uppercase, also I'd say initial at least enough to fill the vertical space (15?) And then new offer batches of 10

imnotteixeira (changes requested):
I still have to review the whole system (the refetching and all) just checked some high level stuff

src/components/HomePage/SearchResultsArea/Offer/OfferItem.js
```diff
@@ -67,7 +67,7 @@ const OfferItem = ({ offer, offerIdx, selectedOfferIdx, setSelectedOfferIdx, loa
                 <ListItemAvatar>
                     {loading ?
                         <Avatar data-testid="avatar-skeleton" variant="circular" classes={{ colorDefault: "transparent" }}>
-                            <Skeleton circle width={100} height={100} />
+                            <Skeleton circle="true" width={100} height={100} />
```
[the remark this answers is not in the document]
TiagooGomess: I made this change because of this new warning that appeared:

image

src/components/HomePage/SearchResultsArea/SearchResultsWidget/useOffersSearcher.js
```diff
@@ -0,0 +1,95 @@
...
+    // In the same way, when we run "setHasMoreOffers(true) on the "search" function,
+    // the "loadMoreOffers" function does not know that the "hasMoreOffers" variable has changed;
+    // Then, when the "loadMoreOffers" function is executed after a previous execution where the
+    // "hasMoreOffers" variable became "false", the state of this variable is still false, which
+    // prevents the fetching of new offers.
+    // Knowing that, I needed a way to set the "hasMoreOffers" variable to "true" when the previous fact
+    // happens: setting the "hasMoreOffers" to true when the "queryToken" (which is stored in redux) changes
+    useEffect(() => {
+        setHasMoreOffers(true);
+    }, [searchQueryToken]);
```
TiagooGomess, line 37: @imnotteixeira I found out that we have a bug when we fetch all the offers of a search and then do another search. What do you think of this way to solve that?

imnotteixeira: I have a solution, but I'm not sure you're going to like it...

Maybe it's time to go full-hook and ditch redux for search functionality.

The main idea is: the HomePage Controller (might need to create it) will start storing the search state by calling the search hook. The search hook will be modified to hold the state and export the setters of each filter and the search value etc. Then it will expose the `search` and `loadMore` functions, as it does currently. It will additionally hold the results array, and errors and loading state, etc... like the redux state currently has.

Then SearchArea and SearchResultsWidget components can both access the HomePage context which will have these methods/data accessible, and most importantly **common** to both. There can only be 1 search at any given time, which is what we want. Components that currently use the redux state for search can easily call `useContext(HomePageContext)` and retrieve whatever they need, no need to pass props around, no connect components to redux.

I'm not sure if we will only ever have the search on the homepage. Maybe some day we will want to keep the search state while the student goes through other pages (viewing offers, adding them to his personal track-list, dunno) and keep the results in a sort-of cart that they can quickly get back to. But for now this is good enough, no need to handle the context above for it to be accessed in other pages.

TiagooGomess: I think it's a good idea

TiagooGomess: @imnotteixeira That refactoring will take some time, mainly due to all the changes in the tests. What do you think of leaving this as it is now (because it works) and opening an issue for that?

imnotteixeira: > @imnotteixeira I found out that we have a bug when we fetch all the offers of a search and then do another search. What do you think of this way to solve that?

Then is this a bug or not?

In any case, we should migrate in the future. Search is the next big system still not using hooks/context. Migrating will probably make it simpler. Create the issue and link it here please

TiagooGomess: It was a bug, but I solved it using that `useEffect`, which uses the `searchQueryToken` from `redux`.

imnotteixeira:
Beware that search service is being changed here #217. Make sure the changes are migrated to this PR as well. Rebases can sometimes hide these changes and the merging goes wrong.

src/components/HomePage/SearchResultsArea/SearchResultsWidget/useOffersSearcher.spec.js
```diff
@@ -0,0 +1,36 @@
...
+        // );
+
+        // expect(notifyHookResult).toHaveBeenCalledWith(expect.objectContaining({
+        //     offers: [],
+        //     hasMoreOffers: false,
+        //     loading: false,
+        //     error: null,
+        // }));
+    });
+});
```
TiagooGomess, line 36: I was reviewing my own PR and found this :eyes:

TiagooGomess: Solved

TiagooGomess:
@imnotteixeira Sorry to bother you, but since you requested changes in this PR, we need your review so that this PR is ready for merging.

PR #173, opened 2022-02-13
Add cookie request snackbar

DoStini:


DoStini:
The current behaviour is the following:
- If the user accepts it, it doesn't prompt again for 15 days
- If the user rejects it, it prompts on reload/entering the site. Im thinking maybe we could add a smaller threshold to this one, for example a day or so. What do you guys think?

src/cookieConsent.js
```diff
@@ -0,0 +1,95 @@
...
+export const CookieConsent = () => {
+    const [open, setOpen] = useState(false);
+
+    const handle = (accepted) => {
+        if (accepted) {
+            initAnalytics();
+        } else {
+            Cookies.remove("_ga");
+            Cookies.remove("_gat");
+            Cookies.remove("_gid");
```
imnotteixeira, line 20: Why do we need an extra package just for this? Just use plain js

https://www.guru99.com/cookies-in-javascript-ultimate-guide.html#:~:text=JavaScript%20Delete%20Cookie,expires%20to%20a%20passed%20date.

src/cookieConsent.js
```diff
@@ -0,0 +1,95 @@
...
+        }
+    };
+
+    useEffect(() => {
+        const cookies = localStorage.getItem("cookies-accepted");
```
imnotteixeira, line 25:
```suggestion
        const cookies = localStorage.getItem("non-essential-cookies-enablement");
```

src/App.js
```diff
@@ -4,12 +4,14 @@ import AppRouter from "./AppRouter";
...
 import Notifier from "./components/Notifications/Notifier";
 import CssBaseline from "@material-ui/core/CssBaseline";
+import { CookieConsent } from "./cookieConsent";
 
 const App = () => (
     <Provider store={store}>
         <React.Fragment>
             <CssBaseline />
             <Notifier />
+            <CookieConsent />
```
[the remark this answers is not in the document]
DoStini: It doesnt

src/cookieConsent.js
DoStini: Removed npm dependency. Do I need to commit the whole Package lock?

imnotteixeira: No, you should untrack package lock from these commits, as no dependency changed

TiagooGomess:
I think we should make the `ACCEPT` option more straightforward than the other. As it's implemented now, I guess the majority of people will not accept the use of all cookies, which is not what we want.

image

After making some small research, I found out that the majority of websites only show the Accept option. For example:



We need to choose between only showing the Accept option, or making the `USE ONLY ESSENTIAL` more "hidden", for example by opening a modal for that option.

DoStini:
I think you might be right. A lot of websites also provide the option to select which cookies to reject, which is a boring process, opposite to this one. What do you think @BrunoRosendo @imnotteixeira

imnotteixeira:
> I think you might be right. A lot of websites also provide the option to select which cookies to reject, which is a boring process, opposite to this one. What do you think @BrunoRosendo @imnotteixeira 

Maybe we could have a "Learn how we use cookies" which opens a modal explaining we use it to know usage patterns. We don't actually know who is doing stuff, and at tge end the button to use only required ones

DoStini:
Yeah, I like that approach

BrunoRosendo:
> I think you might be right. A lot of websites also provide the option to select which cookies to reject, which is a boring process, opposite to this one. What do you think @BrunoRosendo @imnotteixeira

What @imnotteixeira said makes sense to me. I don't want to make the process boring or the pop-up to be visually disturbing, like a lot of websites do. A small pop-up (like it is now) with a redirect button should work well and allows for users to browser the website if they really don't want to accept the cookies.

DoStini:
I don't think we need a different page to explain cookies. I think a small Dialog like Angelo said could do the business. After rejecting it the user could be browsing without any trouble

DoStini:
image

imnotteixeira:
> image
> 

Make the "use only required" button not have the primary color, so that users are drawn to click on the other one instead, also, swap the order: accept on the right

DoStini:
What color do you think will look good? I think the secondary color still calls a lot of attention to the user. It 

image

DoStini:
Maybe a greyish color would be nice.

image

DoStini:
As one last experiment, I think that this one looks good, what do you think?

image

BrunoRosendo:
> Maybe a greyish color would be nice.
> 
> image

I like this one (or the one below it)

imnotteixeira:
> > Maybe a greyish color would be nice.
> > 
> > image
> 
> I like this one (or the one below it)

Don't use this one. Making it gray makes it look like it is disabled which the worst UX. Use black or the secondary

TiagooGomess:
@DoStini Shouldn't the buttons have a background color?

TiagooGomess:
> As one last experiment, I think that this one looks good, what do you think?
> 
> image

Didn't see the background color here, sorry. I think this is the best one.

src/cookieConsent.js
```diff
@@ -0,0 +1,173 @@
+import React, { useEffect, useState } from "react";
+import { Button, Box, Snackbar, Dialog, DialogTitle, DialogContent, DialogActions, makeStyles } from "@material-ui/core";
+import { DAY_IN_MS } from "./utils/TimeUtils";
+import { clearAnalytics, initAnalytics } from "./utils/analytics";
+import PropTypes from "prop-types";
+import { Link } from "react-router-dom";
+import { RouterLink } from "./utils";
+
+export const COOKIE_ACCEPT_LIFETIME_DAYS = 30;
```
TiagooGomess, line 9: Maybe this can be increased to 6 months.

[Source](https://www.cookiebot.com/en/gdpr-cookies/#:~:text=Consent%20must%20be%20renewed%20at,frequent%20renewal%2C%20e.g.%206%20months.):
cookies

BrunoRosendo:
I notices two things:
- Shouldn't the accept button in the pop-up also be on right?
- I can't close the `Why do we use cookies?` window. Is that intended?

DoStini:
> 

Right, what would we do if the user closed the window? Open again the snackbar?

BrunoRosendo:
> > 
> 
> Right, what would we do if the user closed the window? Open again the snackbar?

I think so

TiagooGomess:
Also, "Learn how we use cookies" should have a different color. Another possibility is to change the color of the "Accept" instead. But the "Accept" should be more flashy.

TiagooGomess:
You can also change the background color instead of the colors.

DoStini:
What do you think? @TiagooGomess 

image

imnotteixeira:
Don't forget to test how it looks in mobile

TiagooGomess:
> What do you think? @TiagooGomess The secondary color looks really ugly with this background
> 
> image
> 

I approve

imnotteixeira:
The mobile version is weird, looks like there's a card on top of the notification (look at the shadow). It should work just like a normal notification (snackbar)

DoStini:
> 
That's probably from the MUI snackbar itself, If I'm understanding you correctly, other notifications also have this shadow

image

imnotteixeira:
> > 
> 
> That's probably from the MUI snackbar itself, If I'm understanding you correctly, other notifications also have this shadow
> 
> image

I was referring to the dark grey background (outside of the snackbar). I realized that it is our DevTools (didn't remember they rendered on netlify...)

In that case it's fine.

src/cookieConsent.js
```diff
@@ -0,0 +1,191 @@
...
+                <p>
+                    {"NIJobs uses cookies to collect usage to improve our product."}
+                </p>
+                <p>
+                    {"Regarding privacy issues, we simply collect search parameters and other \
+                        useful metrics without linking them to your identity."}
+                </p>
+                <p>
+                    {"For more information, consult our "}
+                    <Link component={RouterLink} to="privacy">
```
imnotteixeira, line 53:
```suggestion
                    <Link component={RouterLink} to="/privacy-policy">
```

src/cookieConsent.js
```diff
@@ -0,0 +1,191 @@
...
+        });
+        handle(true);
+        setClosed();
+        setClosedWhyModal();
+    };
+
+    const handleReject = () => {
+        const expires = Date.now() + (COOKIE_REJECT_LIFETIME_DAYS * DAY_IN_MS);
+        localStorage.setItem("non-essential-cookies-enablement", {
+            "accepted": false,
```
imnotteixeira, line 118:
```suggestion
            accepted: false,
```

src/cookieConsent.js
```diff
@@ -0,0 +1,191 @@
...
+            handle(cookies.accepted);
+        } else {
+            toggleOpen();
+        }
+    }, [toggleOpen]);
+
+    const handleAccept = () => {
+        const expires = Date.now() + (COOKIE_ACCEPT_LIFETIME_MONTHS * MONTH_IN_MS);
+        localStorage.setItem("non-essential-cookies-enablement", {
+            "accepted": true,
```
imnotteixeira, line 107:
```suggestion
            accepted: true,
```

src/cookieConsent.js
```diff
@@ -0,0 +1,191 @@
...
+                        <Button color="primary" variant="contained" onClick={handleAccept}>Accept</Button>
+                    </Grid>
+                </Grid>
+            </DialogActions>
+        </Dialog>
+    );
+};
+
+export const CookieConsent = () => {
+    const [open, toggleOpen, setClosed] = useToggle(false);
```
imnotteixeira, line 77: This is more consistent with the `useToggle` behavior. If we had a default value of `true` it would do the opposite.

```suggestion
    const [open, toggleOpen, resetConsentOpeningState] = useToggle(false);
```

src/cookieConsent.js
```diff
@@ -0,0 +1,191 @@
...
+
+    const [showWhyModal, toggleWhyModal, setClosedWhyModal] = useToggle(false);
```
imnotteixeira, line 79: This is more consistent with the `useToggle` behavior. If we had a default value of `true` it would do the opposite.


```suggestion
    const [showWhyModal, toggleWhyModal, resetWhyModalOpeningState] = useToggle(false);
```

imnotteixeira (changes requested):

src/cookieConsent.js
```diff
@@ -0,0 +1,190 @@
...
+        if (!expired) {
+            handle(cookies.accepted);
+        } else {
+            toggleOpen();
+        }
+    }, [toggleOpen]);
+
+    const handleAccept = () => {
+        const expires = Date.now() + (COOKIE_ACCEPT_LIFETIME_MONTHS * MONTH_IN_MS);
+        localStorage.setItem("non-essential-cookies-enablement", {
```
imnotteixeira, line 105: `setItem` allows only strings both in key and value [0]. This means that the value is currently being converted to string (JS automatically calls `.toString` on the object you pass as value, resulting in `[object Object]` as you can confirm by inspecting the localStorage after accepting cookies.

You need to `JSON.stringify` the object when setting, and `JSON.parse` when getting the values (in every get and set you have here, respectively).

---

[0] https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #178, opened 2022-02-15
Fix unclosable notification - refactor to Notifier

### LINKED ISSUE #147: Fix un-closeable "Complete your registration" notifications
Currently, the notifications from the `ApplicationMessageNotifier` are not closeable:

https://github.com/NIAEFEUP/nijobs-fe/blob/5b74eefc0cbb6e3ae8b046ed18cd377d622f3cd4/src/components/ApplicationMessages/index.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L9

The issue is that the handleClose handler is not being passed to the Notification ([here](https://github.com/NIAEFEUP/nijobs-fe/blob/5b74eefc0cbb6e3ae8b046ed18cd377d622f3cd4/src/components/ApplicationMessages/index.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L27))

There are many ways to fix this. But looking at our Notifications system implementation, we already had a "default" close handler defined in the [Notifier](https://github.com/NIAEFEUP/nijobs-fe/blob/5b74eefc0cbb6e3ae8b046ed18cd377d622f3cd4/src/components/Notifications/Notifier.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L32). The issue is that it is only used when we call the `addSnackbar` function with the default content, which is not the case in the `ApplicationMessageNotifier`.

---

To fix, we should refactor the Notifier to take notification content options as parameters, and have defaults for the non-required ones. This way, when we called addSanckbar, we could pass contentOptions to tweak the behavior, without having to pass the whole content object (The Notification would automatically be rendered by the Notifier, we would only pass contentOptions when calling `addSanckbar` if needed)

We would need to update other calls which use the content field as well when sending notifications. I only found this one: https://github.com/NIAEFEUP/nijobs-fe/blob/5b74eefc0cbb6e3ae8b046ed18cd377d622f3cd4/src/utils/UndoableActionsHandlerProvider.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L59

BrunoRosendo:
closes #147

### DIFF
--- a/src/components/Notifications/Notifier.js
+++ b/src/components/Notifications/Notifier.js
@@ -19,7 +19,7 @@ const Notifier = ({ notifications, removeSnackbar }) => {
     };
 
     React.useEffect(() => {
-        notifications.forEach(({ key, message, options = {}, dismissed = false }) => {
+        notifications.forEach(({ key, message, options = {}, contentOptions = {}, dismissed = false }) => {
             if (dismissed) {
                 // dismiss snackbar using notistack
                 closeSnackbar(key);
@@ -29,13 +29,14 @@ const Notifier = ({ notifications, removeSnackbar }) => {
             // do nothing if snackbar is already displayed
             if (displayed.current.includes(key)) return;
 
-            const defaultContent = (key, message) =>
+            const configuredContent = (key, message) =>
                 <Notification
-                    message={message}
-                    handleClose={() => closeSnackbar(key)}
+                    {...contentOptions}
+                    message={contentOptions.message || message}
+                    handleClose={contentOptions.handleClose || (() => closeSnackbar(key))}
                 />;
 
-            if (!options.content) options.content = defaultContent;
+            if (!options.content) options.content = configuredContent;
             // display snackbar using notistack
             enqueueSnackbar(message, {
                 key,
--- a/src/utils/UndoableActionsHandlerProvider.js
+++ b/src/utils/UndoableActionsHandlerProvider.js
@@ -3,7 +3,6 @@ import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import { addSnackbar, closeSnackbar } from "../actions/notificationActions";
 import { useTimeout } from "../hooks/useTimeout";
-import Notification from "../components/Notifications/Notification";
 
 const initialActions = {};
 export const UndoableActions = React.createContext(initialActions);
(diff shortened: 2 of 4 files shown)


src/utils/UndoableActionsHandlerProvider.js
```diff
@@ -55,16 +54,13 @@ const BaseActionNotification = ({ action, removeAction, closeSnackbar, addSnackb
...
-                        actionHandler={handleCancel(key)}
-                        handleClose={handleClose(key)}
-                        onMouseEnter={handleMouseEnter}
-                        onMouseLeave={handleMouseLeave}
-                    />,
+            },
+            contentOptions: {
+                actionText: "Undo",
+                actionHandler: handleCancel(action.id),
+                handleClose: handleClose(action.id),
```
BrunoRosendo, line 61: Is passing `action.id` fine? This is not a raw prop anymore so it doesn't recognize `key`

imnotteixeira: I'm not following, could you elaborate?

src/components/Notifications/Notifier.js
```diff
@@ -19,7 +19,7 @@ const Notifier = ({ notifications, removeSnackbar }) => {
     };
 
     React.useEffect(() => {
-        notifications.forEach(({ key, message, options = {}, dismissed = false }) => {
+        notifications.forEach(({ key, message, options = {}, contentOptions = {}, dismissed = false }) => {
```
imnotteixeira, line 22: Add a comment to the addSnackbar function, mentioning the possible parameters, and especially the fact that `options` has precedence over `contentOptions`. If both are present and `options.content` is set, `contentOptions` will not be used

imnotteixeira (changes requested):

src/utils/UndoableActionsHandlerProvider.js
BrunoRosendo: We already spoke about this in the meeting but I had to specify `action.id` since this is not a raw prop anymore

DoStini (approved):

imnotteixeira (approved):

PR #180, opened 2022-02-17
Offer visit metrics

### LINKED ISSUE #175: Metrics - Company and Offer names
In order to filter by company and check how each of its offers is doing, we need to create 2 custom dimensions with the company and offer names, and send them with the offer's pageview (the pageview is already being sent globally but you need to set the dimensions beforehand).

This will allow us, for example, to inform the companies of their offers' popularity.

To better understand the use of custom dimensions, please check #163. This is partly done in the code and partly configured in GA (contact me or @DoStini for access).

BrunoRosendo:
closes #175 

This PR creates metrics for offer visits. Now, we can see how a company is doing by checking how many visits their offers have.
If you want to test this out in GA, you can go to `Behavior->Site Content->All Pages` and filter by the `Offer Visit - Title` custom dimension. You can also create a segment (at the top) to only show a given company's offers (with the Offer Visit - Company` dimension)


src/utils/analytics/index.js
```diff
@@ -107,86 +109,20 @@ export const parseSearchUrl = (queryUrl) => {
...
-    fields: "dimension4",
-    technologies: "dimension5",
-});
+export const recordOfferVisit = (offerId, offerTitle, companyName) => {
+    if (!offerId) return; // Invalid offer, do not set dimensions
+
+    const dimensions = {};
+    dimensions[DIMENSION_IDS.companyName] = companyName;
+    dimensions[DIMENSION_IDS.offerTitle] = offerTitle;
+    ReactGa.set(dimensions);
```
imnotteixeira, line 125:
```suggestion
    ReactGa.set({
[DIMENSION_IDS.companyName]: companyName,
[DIMENSION_IDS.offerTitle]: offerTitle
});
```

src/components/HomePage/SearchResultsArea/Offer/OfferWidget.js
```diff
@@ -45,6 +47,12 @@ const OfferWidget = ({
         });
     }, [isHiddenOffer, offerHiddenReason]);
 
+    useEffect(() => {
+        if (offer) {
+            recordOfferVisit(offer._id, offer.title, offer.ownerName);
+        }
+    }, [offer]);
```
imnotteixeira, line 54: Is this used in both the search result and the offer page? I don't recall.

Also, can you confirm this will only execute when the offer is opened (and not when the search result simply exists)? Just add some console.logs to make sure

Btw, this would be a nice metric as well: how many times an offer has shown in the sewrch results (this is called impressions)

Then we would be able to compare "times shown in results vs. times opened" for any offer/company and check the most "interesting" companies/offers

imnotteixeira (changes requested):

src/utils/analytics/index.js
BrunoRosendo: That's exactly what I wanted but didn't know how :p

src/components/HomePage/SearchResultsArea/Offer/OfferWidget.js
BrunoRosendo: It's used in both the search result and the offer page but I'm getting hits on other pages for some reasons. I'll investigate and see if I need to change this to something else.

That metric would useful! I'll create an issue for it

BrunoRosendo: The problem was that I was using a `user` scope for the GA dimensions. I thought it meant that it'd only hit once per user but instead it hits every pageview for each user. I switched the scope to `hit` but I can only seem to track visits per **session** and not per **device** (user). Is that what we should aim for? Or should we look for another method?

BrunoRosendo: And by the way, the effect only executes when the user selects the offer and not before

imnotteixeira: 1) How is that scope setup done? Is it something we configure on the code, or at GA query time (on the dashboard)? If the latter applies, we need to document that.

2) For the `user` scope, I don't understand why it counts this specific metric for every pageview, since this line only runs on the useEffect.

3) For the `session` scope, does it only count if the user is logged in (I don't think so since that is application logic, and google wouldn't know since we could be using some other cookie for sessions, but still...) If it's only for "browser session", it should be fine, right?

BrunoRosendo: 1. It's only configured in GA, the code is the same. I'll explain these configurations when I get to the docs issue on the BE
2. With the dimension configured with `user` scope, GA will set it for all pageviews from the same user, starting at the one in which the dimension was first set. This is not very useful for a website like ours but it probably has other use cases. You can read more [here](https://support.google.com/analytics/answer/2709828?hl=en#example-user&zippy=%2Cin-this-article)
3. From what I've tested, the session is the browser session (ga doesn't know accounts but tracks each session and user through cookies), so it should be fine. Still, I found it weird to not find the option to filter by user as well, since ga already counts unique users for the website's global visits

imnotteixeira (approved):

PR #193, opened 2022-03-14
Changed the logout button order

### LINKED ISSUE #153: Change the order of the Logout with the admin / company section in the UserMenu
It makes more sense to have the Logout button at the end of the user menu.
You can make the necessary modifications to the code [here](https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/components/Navbar/UserMenu.js#L147).

image

dfneves19:
Changed the logout button order so it appears last.
closes #153

### DIFF
--- a/src/components/Navbar/UserMenu.js
+++ b/src/components/Navbar/UserMenu.js
@@ -55,6 +55,14 @@ const useStyles = makeStyles((theme) => ({
     dividerLabel: {
         padding: theme.spacing(1, 2),
     },
+    menuDivider: {
+        "&&": {
+            margin: theme.spacing(1),
+            marginTop: theme.spacing(1),
+            marginBottom: theme.spacing(1),
+            width: "80%",
+        },
+    }
 }));
 
 const AdminMenuOptions = ({ isMobile }) => {
@@ -152,6 +160,7 @@ const UserMenuContent = React.forwardRef(({ open, isMobile = false, sessionData,
             >
                 {sessionData?.email}
             </Typography>
+            {sessionData?.company && <Divider className={classes.divider} />}
             <MenuList
                 className={classes.menuList}
                 autoFocusItem={open}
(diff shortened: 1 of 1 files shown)


TiagooGomess (approved):

TiagooGomess:
You should rebase with develop before merging

src/components/Navbar/UserMenu.js
```diff
@@ -152,15 +152,15 @@ const UserMenuContent = React.forwardRef(({ open, isMobile = false, sessionData,
             >
                 {sessionData?.email}
             </Typography>
```
TiagooGomess, line 154:
```suggestion
            </Typography>
            {sessionData?.company && <Divider className={classes.divider} />}
```

TiagooGomess: I'm sorry for changing my review; I only saw this after approving the PR.

TiagooGomess (changes requested):

DoStini:
The divider is not yet fixed isn't it? @dfneves19 

image

src/components/Navbar/UserMenu.js
```diff
@@ -152,15 +152,15 @@ const UserMenuContent = React.forwardRef(({ open, isMobile = false, sessionData,
...
-            <Divider className={classes.divider} />
             <MenuList
                 className={classes.menuList}
                 autoFocusItem={open}
                 id="menu-list-grow"
             >
                 {sessionData?.company && <CompanyMenuOptions isMobile={isMobile} sessionData={sessionData} />}
-                <MenuItem button disableTouchRipple onClick={handleLogout}>Logout</MenuItem>
                 {sessionData?.isAdmin && <AdminMenuOptions isMobile={isMobile} />}
+                <Divider className={classes.divider} />
```
DoStini, line 162: In order to fix the divider not being the same width as the top one, I think you need to remove this `classes.divider`. @TiagooGomess

DoStini (changes requested):

dfneves19:
while logged in as a company the dividers are exactly the same, but when logged in as admin the logout divider is still a little deviated to the left.

CarlosMealha:
So I fixed the divider problem with a trivial solution, adding a break statement just to know if it's an admin or a company account authenticated, and then I duplicated the style, just changing the width percentage.  What do you guys think?

Naapperas:
@DoStini please review this PR (since you requested changes previously, also because I would appreciate a second opinion)

CarlosMealha:
> @DoStini please review this PR (since you requested changes previously, also because I would appreciate a second opinion)

I think this would fix the problem
image

Naapperas (approved):
Rebase

BrunoRosendo:
@CarlosMealha This was already approved but it needs to be rebased. Shouldn't take too long

CarlosMealha:
> @CarlosMealha This was already approved but it needs to be rebased. Shouldn't take too long

Okay, it's already done, please aprove it.

PR #194, opened 2022-03-14
Unclear unexpected error in enable offer button in view page as a company 

### LINKED ISSUE #139: Unclear unexpected error in enable offer button in view page as a company
When trying to enable an offer that causes max concurrent offers error, it throws an unexpected error.
It should be clear why the offer was not able to be enabled. This problem will also probably occur when an offer was blocked by an admin

kiko-serra:
Closes #139

### DIFF
--- a/src/components/Company/Registration/Finish/FinishCompanyRegistrationUtils.js
+++ b/src/components/Company/Registration/Finish/FinishCompanyRegistrationUtils.js
@@ -1,4 +1,4 @@
-import { validationRulesGenerator } from "../../../../utils";
+import { validationRulesGenerator, generalHumanError } from "../../../../utils";
 
 export const FinishCompanyRegistrationConstants = {
     bio: {
@@ -20,4 +20,4 @@ const HumanReadableErrors = Object.freeze({
     "registration-already-finished": "This company is already registered.",
 });
 
-export const getHumanError = (error) => HumanReadableErrors[error] || "An error occurred, please try again.";
+export const getHumanError = (error) => generalHumanError(error, HumanReadableErrors);
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -352,3 +352,12 @@ export const buildCancelableRequest = (promiseFn) => (...args) => new Cancelable
 });
 
 export { default as UndoableActionsHandlerProvider, UndoableActions } from "./UndoableActionsHandlerProvider";
+
+export const HumanReadableErrors = Object.freeze({
+    "no-company-found-with-id": (id) => `No company found with given ID: ${id}.`,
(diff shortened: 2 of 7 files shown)


src/components/Company/Registration/Finish/FinishCompanyRegistrationUtils.js
```diff
@@ -20,4 +20,4 @@ const HumanReadableErrors = Object.freeze({
     "registration-already-finished": "This company is already registered.",
 });
 
-export const getHumanError = (error) => HumanReadableErrors[error] || "An error occurred, please try again.";
+export const getHumanError = (error) => generalHumanError(error, HumanReadableErrors);
```
DoStini, line 23: You probably forgot to import `generalHumanError`

src/utils/index.js
```diff
@@ -9,6 +9,7 @@ import useComponentController from "../hooks/useComponentController";
 import CancelablePromise from "cancelable-promise";
 import ReactGa from "react-ga";
 import { OFFER_MAX_LIFETIME_MONTHS } from "./TimeUtils";
+import { string } from "yup";
```
DoStini, line 12: remove.
You should check if the code compiles (which it does not)

src/utils/index.js
```diff
@@ -352,3 +353,14 @@ export const buildCancelableRequest = (promiseFn) => (...args) => new Cancelable
 });
 
 export { default as UndoableActionsHandlerProvider, UndoableActions } from "./UndoableActionsHandlerProvider";
+
+export const generalHumanError = (error, HumanReadableErrors) => {
+    const [errorId, errorValue] = error.split(":");
+    const rawError = HumanReadableErrors[errorId];
+    if (typeof rawError === "string") {
+        return rawError;
+    } else if (typeof rawError === "function"){
```
DoStini, line 362: We'll need to check why your linter is not working

DoStini (changes requested):

DoStini:
also, "all done (i think)" is not a great commit name. Commit names should illustrate what that commit changed. I can help you git rebase to change this

DoStini (changes requested):
Missing here:
https://github.com/NIAEFEUP/nijobs-fe/blob/b672b5a4b3ebfb356b89d8cdd9672ba633586e06/src/components/Apply/Company/CompanyApplicationUtils.js#L26

Also, you're gonna need to rebase and it might be tricky, not sure. Right now we don't have `CreateOfferUtils`, so you need to verify if after the rebase with develop your code is present in this new file `OfferUtils`

DoStini (approved):

PR #223, opened 2022-04-06
feat: add event for impression metrics

limwa:
This pull request adds metrics for when offers appear in search results.

### DIFF
--- a/src/components/HomePage/SearchResultsArea/Offer/OfferItem.js
+++ b/src/components/HomePage/SearchResultsArea/Offer/OfferItem.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import PropTypes from "prop-types";
 
 import {
@@ -55,6 +55,13 @@ const defaultLogo = require("./default_icon.svg");
 const OfferItem = ({ offer, offerIdx, selectedOfferIdx, setSelectedOfferIdx, loading }) => {
     const isCurrentlySelected = selectedOfferIdx === offerIdx;
     const classes = useStyles();
+
+    useEffect(() => {
+        if (offer && !loading) {
+            console.log("Impression on ", offer, { loading });
+        }
+    }, [offer, loading]);
+
     return (
         <div className={classes.itemWrapper}>
             {!loading && <div className={clsx(classes.hoverMask, { [classes.maskActive]: isCurrentlySelected })} />}
--- a/src/components/HomePage/SearchResultsArea/Offer/OfferItem.js
+++ b/src/components/HomePage/SearchResultsArea/Offer/OfferItem.js
@@ -59,7 +59,6 @@ const OfferItem = ({ offer, offerIdx, selectedOfferIdx, setSelectedOfferIdx, loa
 
     useEffect(() => {
         if (offer && !loading) {
-            console.log("Impression on ", offer, { loading });
             recordOfferImpression(offer._id, offer.title, offer.ownerName);
         }
     }, [offer, loading]);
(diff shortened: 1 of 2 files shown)


TiagooGomess:
@limwa, I suggest only doing this after #172, as this is a major change and will break this work.

limwa:
No worries, Bruno told me.
From the code I've looked at #172 should not break this PR though.

limwa:
@BrunoRosendo do you think it is worth it to add a label to the events? If so, what would we put in it?

BrunoRosendo:
> @BrunoRosendo do you think it is worth it to add a label to the events? If so, what would we put in it?

I don't think it's worth it to put a label. It's usually used to add extra information that doesn't belong to the other fields. If we don't have extra information, then it's not needed.

Btw, can you show us how this new event looks in GA? I'm not sure how to compare offers visits with their impressions, for example. Also, can you check if setting these dimensions doesn't trigger unintended visits to the offers? I'm not sure about this.

Honestly, I think the metrics we have are well thought but their implementation/usage is confusing, I hope we're able to improve this in #230, after upgrading to GA4

limwa:
Regarding concerns over this event triggering search events, this was the Google Analytics data for the 4th August, at the start of the test (at midnight, so no events)
image

After open the search area and scrolling down the offers and refreshing a couple times, these were the results (the page is different because the Top Events was not updating correctly)
image

No Offer/visit events were triggered and no pages other than `/` and `/offers` were visited (according to GA), so I think it is safe to say the current implementation doesn't trigger unwanted events or affect other statistics.

Regarding the "how does it look on GA?", to get a representation I go to the "Top Events" page under "Behavior", select "Event action" as the primary dimension and click the Offer/impression action. After that, I select either "Company Name" or "Offer Title" as the secondary dimension and the result is the following
image

You can also generate a Pie chart for this, as such
image

Using Offer Visit - Title as secondary dimension
image

limwa:
Regarding the comparison of impressions vs visits, I'll look into that!

BrunoRosendo:
That's great, thanks a lot for all the steps. We'll have to document it at some point so it's nice to have it written here.

As for the comparison, I don't think it's a problem if you can't put them side by side, we can always look at them separately and compare. I was just wondering if it is possible, it would be very nice

limwa:
I believe that the side-by-side comparison would be easier if we were using events as well for when you click an offer, but I'm not acquainted with GA at all, so I can't be certain of that...

BrunoRosendo:
> I believe that the side-by-side comparison would be easier if we were using events as well for when you click an offer, but I'm not acquainted with GA at all, so I can't be certain of that...

I was thinking about this. Right now, we're using pageviews for offer visits. This mean we can see the unique offer link in the metrics, but quite frankly, it's also a very poor user experience (for us). I would suggest changing the way we do visits and use events in the same way you did in this PR. Additionally, to guarantee we a way of comparing offers with the same title, we could use the offer ID as the event labels (in the visit and the impression). What do you think? @limwa

limwa:
That sounds good! Should I make those changes in this PR?

BrunoRosendo:
> That sounds good! Should I make those changes in this PR?

If you don't mind, I think it'd nice to have it already here

limwa:
Of course! I'll take care of that soon!

BrunoRosendo:
This is looking very good! @limwa 

Do you mind just removing the merge commit? (and while we're at it maybe squash the 2 linting commits into the first commit?)

limwa:
Of course! Just give me one moment!

limwa:
@BrunoRosendo this PR is ready for review

limwa:
As soon as the CI passes :eyes:

BrunoRosendo (approved):
Great job! :rocket:

PR #228, opened 2022-04-09
Added initial config with sentry to report errors

### LINKED ISSUE #226: Add integration with Sentry

monkin77:
closes #226


.env
```diff
@@ -10,4 +10,7 @@ REACT_APP_API_HOSTNAME="https://localhost:8087"
 ANALYTICS_ID=
 
 # GeoLocation Rapid Api key
-GEO_API_KEY="APIKEY"
\ No newline at end of file
+GEO_API_KEY="APIKEY"
+
+# Sentry dsn
+SENTRY_DSN=
```
imnotteixeira, line 16: Add newline at the end of file

imnotteixeira (commented):
What is the added value of Sentry, since we are already posting metrics to google analytics, including the errors?

monkin77:
> What is the added value of Sentry, since we are already posting metrics to google analytics, including the errors?

The main reason was to be notified when the server went down

imnotteixeira:
I'd go with this instead. https://niaefeup.slack.com/apps/A9Z9ZNFE1-uptimebot?tab=more_info 

Adding sentry seems a bit overkill, and it adds up to the dependencies/bundle size, which makes the pages load slower

monkin77:
> I'd go with this instead. https://niaefeup.slack.com/apps/A9Z9ZNFE1-uptimebot?tab=more_info
> 
> Adding sentry seems a bit overkill, and it adds up to the dependencies/bundle size, which makes the pages load slower

sounds good

PR #233, opened 2022-04-09
Feature/disable-offer-visibility

### LINKED ISSUE #218: Disable Offer Visibility Buttons While Fulfilling The Request
Disable the visibility buttons on the offer and the company offers management pages while the respective request is being fulfilled. To test in the Chrome dev tools, go to the Network tab, then choose the throttling option "Slow 3G".

Naapperas:
Closes #218

### DIFF
--- a/src/components/Company/Offers/Manage/CompanyOffersVisibilityActions.js
+++ b/src/components/Company/Offers/Manage/CompanyOffersVisibilityActions.js
@@ -108,7 +108,7 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
 
     return (
         <Tooltip title={offerVisibilityState.isVisible ? "Hide Offer" : "Enable Offer"}>
-            <>
+            <span>
                 <IconButton
                     onClick={handleOfferVisibility}
                     disabled={offerVisibilityButtonDisabled}
@@ -127,7 +127,7 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
                         />
                     }
                 </IconButton>
-            </>
+            </span>
         </Tooltip>
     );
 };
--- a/src/components/Company/Offers/Manage/CompanyOffersVisibilityActions.js
+++ b/src/components/Company/Offers/Manage/CompanyOffersVisibilityActions.js
@@ -29,6 +29,8 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
         isBlocked: undefined,
     });
 
+    const [inMiddleOfRequest, setInMiddleOfRequest] = useState(false);
+
     const isHiddenOffer = offer?.isHidden;
     const offerHiddenReason = offer?.hiddenReason;
 
@@ -40,7 +42,9 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
         });
     }, [isHiddenOffer, offerHiddenReason]);
 
+
     const handleHideOffer = useCallback(({ offer, addSnackbar, onError }) => {
+        setInMiddleOfRequest(true);
         hideOfferService(offer._id)
             .then(() => {
                 setOfferVisibilityState((offerVisibilityState) => ({ ...offerVisibilityState, isVisible: false }));
@@ -51,10 +55,11 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
             })
             .catch((err) => {
                 if (onError) onError(err);
-            });
+            }).finally(() => setInMiddleOfRequest(false));
     }, []);
 
     const handleEnableOffer = useCallback(({ offer, addSnackbar, onError }) => {
+        setInMiddleOfRequest(true);
         enableOfferService(offer._id)
             .then(() => {
                 setOfferVisibilityState((offerVisibilityState) => ({ ...offerVisibilityState, isVisible: true }));
@@ -65,7 +70,7 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
             })
             .catch((err) => {
                 if (onError) onError(err);
-            });
+            }).finally(() => setInMiddleOfRequest(false));
     }, []);
 
     const handleOfferVisibilityError = useCallback((err) => {
@@ -101,13 +106,18 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
 
     return (
         <Tooltip title={offerVisibilityState.isVisible ? "Hide Offer" : "Enable Offer"}>
-            <IconButton
-                onClick={handleOfferVisibility}
-                disabled={offerVisibilityState.isDisabled || offerVisibilityState.isBlocked}
-            >
-                {offerVisibilityState.isVisible ? <VisibilityOffIcon data-testid="HideOffer" color="secondary" fontSize="medium" />
-                    : <VisibilityIcon data-testid="EnableOffer" color="secondary" fontSize="medium" />}
-            </IconButton>
+            <>
+                <IconButton
+                    onClick={handleOfferVisibility}
+                    disabled={inMiddleOfRequest || offerVisibilityState.isDisabled || offerVisibilityState.isBlocked}
+                >
+                    {offerVisibilityState.isVisible ?
+                        <VisibilityOffIcon data-testid="HideOffer" color={!inMiddleOfRequest ? "secondary" : undefined} fontSize="medium" />
+                        :
+                        <VisibilityIcon data-testid="EnableOffer" color={!inMiddleOfRequest ? "secondary" : undefined} fontSize="medium" />
+                    }
+                </IconButton>
+            </>
         </Tooltip>
     );
 };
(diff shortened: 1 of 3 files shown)


TiagooGomess (commented):
Change the name of the state for loadingOfferVisibility

TiagooGomess (changes requested):
I suggest making this change so that the color of the visibility button when the company or offer are blocked also changes:

image

TiagooGomess (commented):
On the offer page, the visibility button does not become disabled while the request is being fulfilled. Can you replicate that on your machine?

Naapperas:
While testing I left a `console.log` in the OfferVIsibilityOptions file and both the "offer details" on the homepage search section and the offer page itself, when pressing the "hid/enable" button, triggered that log, so I assumed that it worked for both, but i might be mistaken.

TiagooGomess:
> While testing I left a `console.log` in the OfferVIsibilityOptions file and both the "offer details" on the homepage search section and the offer page itself, when pressing the "hide/enable" button, triggered that log, so I assumed that it worked for both, but i might be mistaken.

You should enable throttling on the dev tools network tab (for example, slow 3G) so that you can simulate a request that takes some time to be fulfilled. Doing that, can you see the visibility button being disabled?

TiagooGomess:
> Also, should i write tests covering these scenarios ?

You should look at the coverage report and see if there are some uncovered lines (that cover added functionality).

Naapperas:
> > While testing I left a `console.log` in the OfferVIsibilityOptions file and both the "offer details" on the homepage search section and the offer page itself, when pressing the "hide/enable" button, triggered that log, so I assumed that it worked for both, but i might be mistaken.
> 
> You should enable throttling on the dev tools network tab (for example, slow 3G) so that you can simulate a request that takes some time to be fulfilled. Doing that, can you see the visibility button being disabled?

By explicitly waiting 2 seconds (setTimeout) the button appears to be disabled, like this 

image

and it does not fire any events (thus not performing any request to the backend), so we know that the button is disabled at some point. 

What I think is happening is the disabling/enabling is happening faster than the request itself (enabling GPRS or normal 2G, which results in slow requests, didn't seem to make the button get disabled). I added this behavior in 

```js
    const handleOfferVisibility = async () => {
        setLoadingOfferVisibility(true);
        if (visibilityState.isVisible) {
            await handleHideOffer({
                offer: offer,
                addSnackbar: addSnackbar,
                onError: onError,
            });
        } else {
            await handleCompanyEnableOffer({
                offer: offer,
                addSnackbar: addSnackbar,
                onError: onError,
            });
        }
        setLoadingOfferVisibility(false);
    };
```

thinking that the "enabling" part" would only occur after the requests awaited, but it does not seem to be the case. Maybe I could remove the await statement and write this as a then/catch chain, what do you think ?

TiagooGomess:
> > > While testing I left a `console.log` in the OfferVIsibilityOptions file and both the "offer details" on the homepage search section and the offer page itself, when pressing the "hide/enable" button, triggered that log, so I assumed that it worked for both, but i might be mistaken.
> > 
> > 
> > You should enable throttling on the dev tools network tab (for example, slow 3G) so that you can simulate a request that takes some time to be fulfilled. Doing that, can you see the visibility button being disabled?
> 
> By explicitly waiting 2 seconds (setTimeout) the button appears to be disabled, like this
> 
> image
> 
> and it does not fire any events (thus not performing any request to the backend), so we know that the button is disabled at some point.
> 
> What I think is happening is the disabling/enabling is happening faster than the request itself (enabling GPRS or normal 2G, which results in slow requests, didn't seem to make the button get disabled). I added this behavior in
> 
> ```js
>     const handleOfferVisibility = async () => {
>         setLoadingOfferVisibility(true);
>         if (visibilityState.isVisible) {
>             await handleHideOffer({
>                 offer: offer,
>                 addSnackbar: addSnackbar,
>                 onError: onError,
>             });
>         } else {
>             await handleCompanyEnableOffer({
>                 offer: offer,
>                 addSnackbar: addSnackbar,
>                 onError: onError,
>             });
>         }
>         setLoadingOfferVisibility(false);
>     };
> ```
> 
> thinking that the "enabling" part" would only occur after the requests awaited, but it does not seem to be the case. Maybe I could remove the await statement and write this as a then/catch chain, what do you think ?

Yes, you can try that

TiagooGomess:
Not working correctly on my machine:

https://user-images.githubusercontent.com/50210213/162983422-a139dadf-b069-401f-85d0-01cdc3bb1938.mov

TiagooGomess:
When the offer is hidden, the color of the visibility button on the company management offers page is "undefined". It should be secondary. And when the offer is disabled, the color is secondary, but it should be "undefined".

src/components/Company/Offers/Manage/CompanyOffersVisibilityActions.js
```diff
@@ -99,15 +104,30 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
...
     return (
         <Tooltip title={offerVisibilityState.isVisible ? "Hide Offer" : "Enable Offer"}>
-            <IconButton
-                onClick={handleOfferVisibility}
-                disabled={offerVisibilityState.isDisabled || offerVisibilityState.isBlocked}
-            >
-                {offerVisibilityState.isVisible ? <VisibilityOffIcon data-testid="HideOffer" color="secondary" fontSize="medium" />
-                    : <VisibilityIcon data-testid="EnableOffer" color="secondary" fontSize="medium" />}
-            </IconButton>
+            <>
```
TiagooGomess, line 111:
```suggestion
            
```

src/components/Company/Offers/Manage/CompanyOffersVisibilityActions.js
```diff
@@ -99,15 +104,30 @@ const CompanyOffersVisibilityActions = ({ offer }) => {
...
+                        />
+                        :
+                        <VisibilityIcon
+                            data-testid="EnableOffer"
+                            color={offerVisibilityButtonDisabled ? undefined : "secondary"}
+                            fontSize="medium"
+                        />
+                    }
+                </IconButton>
+            </>
```
TiagooGomess, line 130:
```suggestion
            
```

src/components/HomePage/SearchResultsArea/Offer/OfferVisibilityOptions.js
```diff
@@ -75,31 +73,33 @@ const OfferVisibilityOptions = ({
...
-                                onClick={handleOfferVisibility}
-                                className={classes.visibilityButton}
-                                role="hideEnableOfferButton"
-                                disabled={visibilityState.isDisabled}
-                                startIcon={visibilityState.isVisible ? <VisibilityOff /> : <Visibility />}
-                            >
-                                { visibilityState.isVisible ? "Hide Offer" : "Enable Offer" }
-                            </Button>
+                        <Tooltip title={visibilityState.isVisible ? "Hide Offer" : "Enable Offer"}>
+                            <>
```
TiagooGomess, line 80:
```suggestion
                           
```

src/components/HomePage/SearchResultsArea/Offer/OfferVisibilityOptions.js
```diff
@@ -75,31 +73,33 @@ const OfferVisibilityOptions = ({
...
+                                <Button
+                                    onClick={handleOfferVisibility}
+                                    className={classes.visibilityButton}
+                                    role="hideEnableOfferButton"
+                                    disabled={loadingOfferVisibility || visibilityState.isDisabled}
+                                    startIcon={visibilityState.isVisible ? <VisibilityOff /> : <Visibility />}
+                                >
+                                    {visibilityState.isVisible ? "Hide Offer" : "Enable Offer"}
+                                </Button>
+                            </>
```
TiagooGomess, line 90:
```suggestion
                            
```

TiagooGomess (changes requested):
You must remove these; otherwise, the tooltips don't work.

Naapperas:
> You must remove these; otherwise, the tooltips don't work.

I had changed those because MUI would complain about providing a disabled child to the Tooltip component and I figured that a React fragment would work, changed it to a `span` instead.

TiagooGomess:
Do the same for [OfferVisibilityOptions.js](https://github.com/NIAEFEUP/nijobs-fe/pull/233/files/612dd757c7f466c9640a4a38af9d3e40ed44a67a#diff-5004fe71a0fb5f48ecd3f9cbe5301cce777f84e76a9105d0c0e62ffc81127283)

src/services/locationSearchService.js
```diff
@@ -1,27 +1,37 @@
...
-        return data.map(({ city, country, latitude, longitude }) => ({ city, country, latitude, longitude }));
+        // const data = (await res.json()).data;
+        // return data.map(({ city, country, latitude, longitude }) => ({ city, country, latitude, longitude }));
+
+        const data = await res.json();
+
+        return data;
 
     } catch (err) {
         throw [{ msg: Constants.UNEXPECTED_ERROR_MESSAGE }];
```
TiagooGomess, line 37: Maybe you should not modify this file in this PR

Naapperas: I don't now how this got through...

TiagooGomess (approved):

PR #244, opened 2022-05-25
NIJobs version with infinite scrolling

BrunoRosendo:


TiagooGomess (approved):
Already tested on the beta version and it's working fine.

PR #248, opened 2022-06-22
Feature/change log

### LINKED ISSUE #245: Create change log page
- Investigate what info the github API can fetch
- Try to get the information from github releases

The link to this new page could be here
image

monkin77:
Closes #245

### DIFF
--- a/src/AppRouter.js
+++ b/src/AppRouter.js
@@ -219,10 +219,10 @@ const AppRouter = () => (
                     <CreateOfferPage />
                 </PageLayout>
             </Route>
-            <Route exact path="/whatsNew">
+            <Route exact path="/whats-new">
                 <PageLayout
-                    key="/changeLog"
-                    pageTitle="Changelog"
+                    key="/whats-new"
+                    pageTitle="What's new?"
                     layout={LayoutType.DESKTOP}
                 >
                     <ChangeLogPage />
--- a/src/AppRouter.js
+++ b/src/AppRouter.js
@@ -11,7 +11,10 @@ import CompanyApplicationPage, {
 import ApplicationsReviewPage from "./pages/ApplicationsReviewPage";
 import NotFound from "./pages/NotFound";
 import ErrorPage from "./pages/ErrorPage";
-import OfferPage, { OfferPageController, OfferPageControllerContext } from "./pages/OfferPage";
+import OfferPage, {
+    OfferPageController,
+    OfferPageControllerContext,
+} from "./pages/OfferPage";
 import RulesPage from "./pages/RulesPage";
 import { ProtectedRoute, Route } from "./utils";
 import PageLayout, { LayoutType } from "./components/Layout/PageLayout";
@@ -22,9 +25,15 @@ import {
 import FinishCompanyRegistrationPage from "./pages/FinishCompanyRegistrationPage";
 import CompanyOffersManagementPage from "./pages/CompanyOffersManagementPage";
 import CreateOfferPage from "./pages/CreateOfferPage";
-import { CreateOfferController, CreateOfferControllerContext } from "./components/Offers/New/CreateOfferForm";
+import {
+    CreateOfferController,
+    CreateOfferControllerContext,
+} from "./components/Offers/New/CreateOfferForm";
 import { CookieConsent } from "./cookieConsent";
-import { EditOfferController, EditOfferControllerContext } from "./components/Offers/Edit/EditOfferForm";
+import {
+    EditOfferController,
+    EditOfferControllerContext,
+} from "./components/Offers/Edit/EditOfferForm";
 import EditOfferPage from "./pages/EditOfferPage";
 import PrivacyPolicyPage from "./pages/PrivacyPolicyPage";
 import TermsAndConditionsPage from "./pages/TermsAndConditionsPage";
@@ -40,17 +49,16 @@ import ChangeLogPage from "./pages/ChangeLogPage";
  *
  */
 
-const shouldShowCompanyApplicationMobile = ({ showConfirmationModal, isMobileSize }) => !showConfirmationModal && isMobileSize;
+const shouldShowCompanyApplicationMobile = ({
+    showConfirmationModal,
+    isMobileSize,
+}) => !showConfirmationModal && isMobileSize;
 
 const AppRouter = () => (
     <BrowserRouter basename={`${process.env.REACT_APP_BASE_ROUTE || "/"}`}>
         <CookieConsent />
         <Switch>
-            <Route
-                exact
-                path="/"
-                key="/"
-            >
+            <Route exact path="/" key="/">
                 <PageLayout
                     key="/"
                     showHomePageLink={false}
@@ -60,11 +68,7 @@ const AppRouter = () => (
                     <HomePage />
                 </PageLayout>
             </Route>
-            <Route
-                exact
-                path="/privacy-policy"
-                key="/privacy-policy"
-            >
+            <Route exact path="/privacy-policy" key="/privacy-policy">
                 <PageLayout
                     key="/privacy-policy"
                     pageTitle="Privacy Policy"
@@ -104,10 +108,7 @@ const AppRouter = () => (
                     <CompanyApplicationPage />
                 </PageLayout>
             </Route>
-            <Route
-                exact
-                path="/rules"
-            >
+            <Route exact path="/rules">
                 <PageLayout
                     key="/rules"
                     pageTitle="Rules"
@@ -122,7 +123,7 @@ const AppRouter = () => (
                 path="/review/applications"
                 unauthorizedRedirectPath="/"
                 unauthorizedRedirectMessage="You are not allowed to access the applications review page."
-                authorize={(user) => (user.isAdmin)}
+                authorize={(user) => user.isAdmin}
             >
                 <PageLayout
                     key="/review/applications"
@@ -138,7 +139,9 @@ const AppRouter = () => (
                 path="/company/registration/finish"
                 unauthorizedRedirectPath="/"
                 unauthorizedRedirectMessage="To access this page you must be logged in and have a pending registration."
-                authorize={(user) => (user.company && !user.company.hasFinishedRegistration)}
+                authorize={(user) =>
+                    user.company && !user.company.hasFinishedRegistration
+                }
                 context={FinishCompanyRegistrationControllerContext}
                 controller={FinishCompanyRegistrationController}
             >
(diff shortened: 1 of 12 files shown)


monkin77:
changelog1
changelog2

BrunoRosendo:
Some feedback based only by looking at the page:
- The symbol besides `Latest Changes` feels a bit weird to me. Maybe we could find one that suits better?
- The page's title is `Changelog` but the tab/window title is `Change Log`, we should make them equal. I would even consider something like `What's new?`. What do you think?
- The page is very similar to the rules/terms and conditions pages but its layout and title are different. I think we should standardize this. You can choose which one you prefer
- I feel like the title of each release is a bit small. Maybe try to increase it a bit? (something like the subtitles of the terms and conditions page)

A side note, the rules page's title seems to be broken. If it's easy to fix, could you just do it here? Otherwise, we can create an issue

src/pages/ChangeLogPage.js
```diff
@@ -0,0 +1,132 @@
...
+                if (lastTS !== null)
+                    setLastUpdateTS(format(lastTS, "dd-MM-yyyy"));
+                return filteredReleases;
+            });
+            setReleases(releases);
+        };
+
+        fetchData().catch((_) => {
+            // setError(err[0]);
+            // setIsLoading(false);
```
BrunoRosendo, line 71: I think this can be removed

src/pages/ChangeLogPage.js
```diff
@@ -0,0 +1,132 @@
...
+                });
+
+                if (lastTS !== null)
+                    setLastUpdateTS(format(lastTS, "dd-MM-yyyy"));
+                return filteredReleases;
+            });
+            setReleases(releases);
+        };
+
+        fetchData().catch((_) => {
```
BrunoRosendo, line 69: Is there a reason to use then/catch here instead of the `await` logic? Maybe it would be cleaner

src/pages/ChangeLogPage.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        );
+
+        expect(wrapper.getByText("Changelog")).toBeInTheDocument();
+
+        await waitFor(() => {
+            expect(addSnackbar).toHaveBeenCalledTimes(1);
+        }, {
+            timeout: 1000,
+        });
+    });
```
BrunoRosendo, line 127: I know this is also used somewhere else but why is the timeout needed for this snackbar but not for the cookie notification, for example? Just want to understand the difference

src/services/changeLogService.js
```diff
@@ -0,0 +1,27 @@
...
+
+        return json;
+    } catch (error) {
+        const errorArray = Array.isArray(error)
+            ? error
+            : [{ msg: Constants.UNEXPECTED_ERROR_MESSAGE }];
+
+        throw errorArray;
+    }
+});
```
BrunoRosendo, line 27: I think we should record analytics about this. You can follow other services and add `measureTime` and create error events

BrunoRosendo (changes requested):

BrunoRosendo:
This is going really well until now, good job!

monkin77:
> Some feedback based only on looking at the page:
> 
> * The symbol besides `Latest Changes` feels a bit weird to me. Maybe we could find one that suits better?
> * The page's title is `Changelog` but the tab/window title is `Change Log`, we should make them the same. I would even consider something like `What's new?`. What do you think?
> * The page is very similar to the rules/terms and conditions pages but its layout and title are different. I think we should standardize this. You can choose which one you prefer
> * I feel like the title of each release is a bit small. Maybe try to increase it a bit? (something like the subtitles of the terms and conditions page)
> 
> A side note, the rules page's title seems to be broken. If it's easy to fix, could you just do it here? Otherwise, we can create an issue

Regarding the icon, I'm not sure which one would fit better in this case. Could be discussed in the next meeting.

Regarding the Title, I've changed it to be equal to the page title and I think the change towards "What's new?" could also be discussed at the meeting.

I'll take care of the other changes for now.

src/pages/ChangeLogPage.spec.js
monkin77: I'm not aware of the cookie notification case, but here I use this waitFor since the addSnackbar method is being called after a certain method (fetchData) catches an error, so it is not called instantly. I'm not 100% sure this is the reason why, but using the `act` method to wait for updates is not enough for this case, I think it is because there isn't a component state update, only an error catching. By using the `waitFor` it will assert multiple times and will finish earlier if the assertion is correct.

DoStini:
I think we can use this one: newspaper

image

DoStini:
We think that "What's new" is a cool title and button text

monkin77:
image
What about this icon?

BrunoRosendo:
@monkin77 https://mui.com/blog/material-ui-is-now-mui/
I think this means the @material-ui packages were abandoned and replaced by the @mui ones? For examples, [this is the one we're using](https://www.npmjs.com/package/@material-ui/icons) but [this one has the newspaper icon](https://www.npmjs.com/package/@mui/icons-material). Can you check if we should update our packages for this @mui ones?

I also found this migration guide: https://mui.com/pt/material-ui/migration/migration-v4/
I'm not sure how hard this would be, if it's worth it or whether it should be done here or in a new PR. What do you think?

monkin77:
> @monkin77 https://mui.com/blog/material-ui-is-now-mui/ I think this means the @material-ui packages were abandoned and replaced by the @mui ones? For examples, [this is the one we're using](https://www.npmjs.com/package/@material-ui/icons) but [this one has the newspaper icon](https://www.npmjs.com/package/@mui/icons-material). Can you check if we should update our packages for this @mui ones?
> 
> I also found this migration guide: https://mui.com/pt/material-ui/migration/migration-v4/ I'm not sure how hard this would be, if it's worth it or whether it should be done here or in a new PR. What do you think?

I believe this should be done in another PR, since the new mui library requires the React library to be at least in version 17. We could create an issue to take care of upgrading the React version (either 17 or the most updated) and consequently use the new mUI library?

BrunoRosendo:
> > @monkin77 https://mui.com/blog/material-ui-is-now-mui/ I think this means the @material-ui packages were abandoned and replaced by the @mui ones? For examples, [this is the one we're using](https://www.npmjs.com/package/@material-ui/icons) but [this one has the newspaper icon](https://www.npmjs.com/package/@mui/icons-material). Can you check if we should update our packages for this @mui ones?
> > I also found this migration guide: https://mui.com/pt/material-ui/migration/migration-v4/ I'm not sure how hard this would be, if it's worth it or whether it should be done here or in a new PR. What do you think?
> 
> I believe this should be done in another PR, since the new mui library requires the React library to be at least in version 17. We could create an issue to take care of upgrading the React version (either 17 or the most updated) and consequently use the new mUI library?

Oh that makes it a lot harder then. Can you please create an issue for that and we can discuss it in a meeting?

src/AppRouter.js
```diff
@@ -215,10 +219,16 @@ const AppRouter = () => (
                     <CreateOfferPage />
                 </PageLayout>
             </Route>
-            <Route
-                path="/error"
-                key="/error"
-            >
+            <Route exact path="/whatsNew">
```
BrunoRosendo, line 222: This could be whats-new, to be consistent with privacy-policy and terms-and-conditions

src/AppRouter.js
```diff
@@ -215,10 +219,16 @@ const AppRouter = () => (
...
+                <PageLayout
+                    key="/changeLog"
+                    pageTitle="Changelog"
```
BrunoRosendo, line 225: This is the outdated title

src/components/Layout/ContactSection.js
```diff
@@ -79,68 +93,121 @@ const ContactSection = () => {
...
+                                                }
+                                            />
+                                        }
                                         className={classes.colorText}
                                         title="facebook"
                                     >
-                                        <Typography variant="button">Facebook</Typography>
+                                        <Typography variant="button">
+                                            Facebook
+                                        </Typography>
```
BrunoRosendo, line 126: This is not a change request but there are lot of diffs caused by your auto formatter, it makes the review a bit harder to do. Maybe change its settings so it only triggers in your own changes?

src/pages/ChangeLogPage.js
```diff
@@ -0,0 +1,129 @@
...
+
+    return (
+        <div className={classes.content}>
+            <Typography
+                variant="h3"
+                id="Change-Log"
+                data-id="Change-Log"
+                gutterBottom
+            >
+                What&apos;s new?
```
BrunoRosendo, line 86: Same thing as above

src/pages/ChangeLogPage.js
```diff
@@ -0,0 +1,129 @@
...
+        padding: theme.spacing(2),
+    },
+}));
+
+export const ChangeLogPage = ({ addSnackbar }) => {
+    const isMobile = useMobile();
+    const classes = useStyles({ isMobile: isMobile });
+
+    const [releases, setReleases] = useState([]);
+    const [lastUpdateTS, setLastUpdateTS] = useState(null);
```
BrunoRosendo, line 42: What is `TS`?

BrunoRosendo (changes requested):
Some things
- The tab title is still `Changelog`, it should also be `What's New?`
- The `rules` and `privacy policy` pages have a larger body than this one, I think it should be consistent
- What if we switch this logo with the one that  Rules` uses? Maybe it makes more sense to have this "danger" look

What do you think?

src/components/Layout/ContactSection.js
monkin77: I found a setting for that, thanks :smiley_cat:

src/pages/ChangeLogPage.js
monkin77: Timestamp

BrunoRosendo:
> Some things
> 
>     * The tab title is still `Changelog`, it should also be `What's New?`
> 
>     * The `rules` and `privacy policy` pages have a larger body than this one, I think it should be consistent
> 
>     * What if we switch this logo with the one that  Rules` uses? Maybe it makes more sense to have this "danger" look
> 
> 
> What do you think?

As for my last 2 comments on the list, is this something you could do?

monkin77:
> > Some things
> > ```
> > * The tab title is still `Changelog`, it should also be `What's New?`
> > 
> > * The `rules` and `privacy policy` pages have a larger body than this one, I think it should be consistent
> > 
> > * What if we switch this logo with the one that  Rules` uses? Maybe it makes more sense to have this "danger" look
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think?
> 
> As for my last 2 comments on the list, is this something you could do? @monkin77

Sorry, I forgot about those. Here's how it would look switching the icons:

BrunoRosendo (approved):
Amazing job! Let's go :rocket:

PR #251, opened 2022-07-16
Feature/search url params

### LINKED ISSUE #220: Build url params when clicking in the search

Naapperas:
Closes #220

### DIFF
--- a/src/components/HomePage/MainView.spec.js
+++ b/src/components/HomePage/MainView.spec.js
@@ -7,6 +7,8 @@ import ShowMoreButton from "./ShowMoreButton";
 import { createTheme } from "@material-ui/core/styles";
 import { mountWithStore } from "../../test-utils";
 
+import { MemoryRouter } from "react-router-dom";
+
 describe("Main View", () => {
     let scrollToProductDescription, showSearchResults, wrapper;
     const theme = createTheme({});
@@ -24,10 +26,12 @@ describe("Main View", () => {
         showSearchResults = jest.fn();
 
         wrapper = mountWithStore(
-            <MainView
-                scrollToProductDescription={scrollToProductDescription}
-                showSearchResults={showSearchResults}
-            />,
+            <MemoryRouter initialEntries={["/"]}>
+                <MainView
+                    scrollToProductDescription={scrollToProductDescription}
+                    showSearchResults={showSearchResults}
+                />
+            </MemoryRouter>,
             initialState,
             theme
         );
@@ -59,7 +63,7 @@ describe("Main View", () => {
             fetch.mockResponse(JSON.stringify({ mockData: true }));
 
             wrapper.find("form#search_form").first().simulate("submit", {
-                preventDefault: () => {},
+                preventDefault: () => { },
             });
             expect(showSearchResults).toHaveBeenCalledTimes(1);
         });
--- a/src/components/HomePage/SearchArea/SearchArea.js
+++ b/src/components/HomePage/SearchArea/SearchArea.js
@@ -1,6 +1,9 @@
 import React, { useCallback } from "react";
 import PropTypes from "prop-types";
 
+import { useLocation, useHistory } from "react-router-dom";
+import qs from "qs";
+
 import { connect } from "react-redux";
 import { SearchResultsConstants } from "../SearchResultsArea/SearchResultsWidget/SearchResultsUtils";
 import {
@@ -27,6 +30,8 @@ import AdvancedSearchDesktop from "./AdvancedSearch/AdvancedSearchDesktop";
 import useComponentController from "../../../hooks/useComponentController";
 import useOffersSearcher from "../SearchResultsArea/SearchResultsWidget/useOffersSearcher";
 
+import { throttle } from "lodash";
+
 export const AdvancedSearchControllerContext = React.createContext({});
 
 export const AdvancedSearchController = ({
@@ -35,20 +40,98 @@ export const AdvancedSearchController = ({
     resetAdvancedSearchFields, onSubmit, searchValue, setSearchValue, onMobileClose,
 }) => {
 
+    const location = useLocation();
+    const history = useHistory();
+
+    // since we are always changing the URL (and therefore window.location), there is no point in memoizing this value
+    const queryParams = qs.parse(location.search, {
+        ignoreQueryPrefix: true,
+        arrayFormat: "brackets",
+    });
+
+    // need to throttle down calling 'history.replace' because
+    // too many invocations might cause the browser to stop responding
+    // value of 350 derived from https://github.com/PoorBillyPilgrim/artcrawl-digital-exhibit/issues/20#issue-761626379
+    // according to this 1000/3 is the lower bound for the 'wait' value
+    const changeURLFilters = useCallback(throttle((location, history, queryParams, changes) => {
+        const newQueryParams = {
+            ...queryParams,
+            ...changes,
+        };
+
+        history.replace({
+            ...location, search: qs.stringify(newQueryParams, {
+                skipNulls: true,
+                arrayFormat: "brackets",
+            }),
+        });
+    }, 350), []);
+
+    const clearURLFilters = useCallback(
+        throttle(
+            (location, history) => history.replace({ ...location, search: "" }),
+            350
+        ), []);
+
+    const actualSetJobType = useCallback(({ target: { value: jobType } }) => {
+        changeURLFilters(location, history, queryParams, { jobType });
+
+        setJobType(jobType);
+    }, [changeURLFilters, location, history, queryParams, setJobType]);
+
+    const actualSetJobDuration = useCallback((unused, duration) => {
+
+        const [jobMinDuration, jobMaxDuration] = duration;
+
+        changeURLFilters(location, history, queryParams, { jobMinDuration, jobMaxDuration });
+
+        setJobDuration(unused, duration);
+    }, [changeURLFilters, location, history, queryParams, setJobDuration]);
+
+    const actualSetShowJobDurationSlider = useCallback((showJobDurationSlider) => {
+        if (!showJobDurationSlider) {
+            changeURLFilters(location, history, queryParams, { jobMinDuration: null, jobMaxDuration: null });
+
+            setJobDuration(null, [null, null]);
+        }
+
+        setShowJobDurationSlider(showJobDurationSlider);
+    }, [changeURLFilters, history, location, queryParams, setJobDuration, setShowJobDurationSlider]);
+
+    const actualSetFields = useCallback((fields) => {
+
+        changeURLFilters(location, history, queryParams, { fields });
+
+        setFields(fields);
+    }, [changeURLFilters, history, location, queryParams, setFields]);
+
+    const actualSetTechs = useCallback((technologies) => {
+
+        changeURLFilters(location, history, queryParams, { technologies });
+
+        setTechs(technologies);
+    }, [changeURLFilters, history, location, queryParams, setTechs]);
+
+    const actualResetAdvancedSearchFields = useCallback(() => {
+        clearURLFilters(location, history);
+
+        resetAdvancedSearchFields();
+    }, [clearURLFilters, history, location, resetAdvancedSearchFields]);
+
     const advancedSearchProps = useAdvancedSearch({
         enableAdvancedSearchDefault,
-        jobMinDuration,
-        jobMaxDuration,
-        setJobDuration,
-        showJobDurationSlider,
-        setShowJobDurationSlider,
-        jobType,
-        setJobType,
-        fields,
-        setFields,
-        technologies,
-        setTechs,
-        resetAdvancedSearchFields,
+        jobMinDuration, // TODO: make it so this fields reflect the respective duration
+        jobMaxDuration, // TODO: make it so this fields reflect the respective duration
+        setJobDuration: actualSetJobDuration,
+        showJobDurationSlider: showJobDurationSlider, // TODO: make so this reflects weathere there are job durations or not
+        setShowJobDurationSlider: actualSetShowJobDurationSlider,
+        jobType: queryParams.jobType ?? jobType,
+        setJobType: actualSetJobType,
+        fields: queryParams.fields ?? fields,
+        setFields: actualSetFields,
+        technologies: queryParams.technologies ?? technologies,
+        setTechs: actualSetTechs,
+        resetAdvancedSearchFields: actualResetAdvancedSearchFields,
     });
 
     const { search: searchOffers } = useOffersSearcher({
@@ -64,8 +147,10 @@ export const AdvancedSearchController = ({
         if (e) e.preventDefault();
         searchOffers(SearchResultsConstants.INITIAL_LIMIT);
 
+        changeURLFilters(location, history, queryParams, { searchValue });
+
         if (onSubmit) onSubmit();
-    }, [onSubmit, searchOffers]);
+    }, [changeURLFilters, history, location, onSubmit, queryParams, searchOffers, searchValue]);
 
     return {
         ...advancedSearchProps,
@@ -177,7 +262,7 @@ export const mapStateToProps = ({ offerSearch }) => ({
 export const mapDispatchToProps = (dispatch) => ({
     setSearchValue: (value) => dispatch(setSearchValue(value)),
     setJobDuration: (_, value) => dispatch(setJobDuration(...value)),
-    setJobType: (e) => dispatch(setJobType(e.target.value)),
+    setJobType: (jobType) => dispatch(setJobType(jobType)),
     setFields: (fields) => dispatch(setFields(fields)),
     setTechs: (technologies) => dispatch(setTechs(technologies)),
     setShowJobDurationSlider: (val) => dispatch(setShowJobDurationSlider(val)),
--- a/src/components/HomePage/SearchArea/SearchArea.js
+++ b/src/components/HomePage/SearchArea/SearchArea.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo } from "react";
+import React, { useCallback, useEffect, useMemo, useRef } from "react";
 import PropTypes from "prop-types";
 
 import { useLocation, useHistory } from "react-router-dom";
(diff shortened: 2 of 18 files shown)


Naapperas:
I'm having a problem where, due to the use of React Contexts, the `useLocation` and `useHistory` hooks are throwing errors inside the testing environment.

I have already tried mocking the `react-router-dom` and `react-router` modules but unsuccessfully.

Here are several links all regarding the same issue (`react-router` hooks not working):
[Link 1](https://stackoverflow.com/questions/58117890/how-to-test-components-using-new-react-router-hooks/58206121#58206121)
[Link 2](https://stackoverflow.com/questions/58117890/how-to-test-components-using-new-react-router-hooks?answertab=scoredesc#tab-top)
[Link 3](https://stackoverflow.com/a/60536525)
[Link 4](https://v5.reactrouter.com/web/guides/testing/checking-location-in-tests)

The best solution, according to the links provided, to solve this issue is to, in every custom `mount` and `render` setup [here](https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/test-utils.js), wrap the children to be rendered/mounted inside a `Router` instance: I have already tested locally using a `MemoryRouter` and in fact it does work. 
However, changing all these methods is a bit cumbersome and I cannot help but think that there might be a more efficient and elegant way to tackle this issue. Any suggestions ? @DoStini @imnotteixeira @TiagooGomess

imnotteixeira:
> I'm having a problem where, due to the use of React Contexts, the `useLocation` and `useHistory` hooks are throwing errors inside the testing environment.
> 
> I have already tried mocking the `react-router-dom` and `react-router` modules but unsuccessfully.
> 
> Here are several links all regarding the same issue (`react-router` hooks not working): [Link 1](https://stackoverflow.com/questions/58117890/how-to-test-components-using-new-react-router-hooks/58206121#58206121) [Link 2](https://stackoverflow.com/questions/58117890/how-to-test-components-using-new-react-router-hooks?answertab=scoredesc#tab-top) [Link 3](https://stackoverflow.com/a/60536525) [Link 4](https://v5.reactrouter.com/web/guides/testing/checking-location-in-tests)
> 
> The best solution, according to the links provided, to solve this issue is to, in every custom `mount` and `render` setup [here](https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/test-utils.js), wrap the children to be rendered/mounted inside a `Router` instance: I have already tested locally using a `MemoryRouter` and in fact it does work. However, changing all these methods is a bit cumbersome and I cannot help but think that there might be a more efficient and elegant way to tackle this issue. Any suggestions ? @DoStini @imnotteixeira @TiagooGomess

This is already what we do. In every component that has a link (which implies having a Router), we wrap it in a Router. I don't think we should wrap **everything** in a router by default, but we should wrap every component that needs it, maybe create a Dummy utility component in your test file to automatically wrap the argument it receives in a Router similar to what we do [here](https://github.com/NIAEFEUP/nijobs-fe/blob/develop/src/components/Company/Registration/Finish/FinishCompanyRegistrationWidget.spec.js#L20-L35)?

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -64,8 +147,35 @@ export const AdvancedSearchController = ({
         if (e) e.preventDefault();
         searchOffers(SearchResultsConstants.INITIAL_LIMIT);
 
+        changeURLFilters(location, history, queryParams, { searchValue: queryParams.searchValue ?? searchValue });
+
         if (onSubmit) onSubmit();
-    }, [onSubmit, searchOffers]);
+    }, [changeURLFilters, history, location, onSubmit, queryParams, searchOffers, searchValue]);
+
+    const firstRender = useRef(true);
```
imnotteixeira, line 155: This is not the way to do stuff only on first renders. React has a built-in mechanism for that, which is to use a useEffect with an empty deps array `useEffect(fn, [])`

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -35,20 +40,98 @@ export const AdvancedSearchController = ({
...
+
+        setTechs(technologies);
+    }, [changeURLFilters, history, location, queryParams, setTechs]);
+
+    const actualResetAdvancedSearchFields = useCallback(() => {
+        clearURLFilters(location, history);
+
+        resetAdvancedSearchFields();
+    }, [clearURLFilters, history, location, resetAdvancedSearchFields]);
+
```
imnotteixeira, line 120: This is too verbose to be within the component code, it may be better to extract to a custom hook of ours, and just call the hook here and make it return whatever data we need

src/components/HomePage/SearchArea/SearchArea.js
Naapperas: I had tested with that option but at the time it was not behaving as I expected: maybe I did something else in a way that produced wrong results and which led me to that solution. Tested the change now and behaves as intended.

Naapperas:
I am still missing the "auto-submit" feature when the search term is already present in the URL but other than that the PR is ready.

Naapperas:
I came up with the solution of using a global boolean to check weather we were on the first ever render or not (relying on mounting alone, like with `useEffect(..., [])` does not work because apparently the component gets re-mounted after a submit, which would lead to infinite requests) but this seems like black magic. However, i have not been able to find a reasonable solution to this: any ideas ?

Naapperas:
image
image

This was after clicking on the search button multiple times

imnotteixeira:
> I came up with the solution of using a global boolean to check weather we were on the first ever render or not (relying on mounting alone, like with `useEffect(..., [])`, does not work because apparently the component gets re-mounted after a submit, which would lead to infinite requests) but this seems like black magic. However, i have not been able to find a reasonable solution to this: any ideas ?

Yes! Welcome to React :)

That's a nice feature that hints you to not put this behavior in that component: why simulate a form submission if the user hasn't submitted anything? We can still call our own "redux-linked" search function elsewhere (I'd say on HomePage level, next to the MainView, so that we can toggle the showSearchResults variable easily. That component will not re-render and we avoid the multiple calls.

This goes in pair to what I've been suggesting for a while: to have the search system accessible by a hook, and independent from the components that use it (SearchAreas and such), but that is to be done in a separate issue

Naapperas:
> > I came up with the solution of using a global boolean to check weather we were on the first ever render or not (relying on mounting alone, like with `useEffect(..., [])`, does not work because apparently the component gets re-mounted after a submit, which would lead to infinite requests) but this seems like black magic. However, i have not been able to find a reasonable solution to this: any ideas ?
> 
> Yes! Welcome to React :)
> 
> That's a nice feature that hints you to not put this behavior in that component: why simulate a form submission if the user hasn't submitted anything? We can still call our own "redux-linked" search function elsewhere (I'd say on HomePage level, next to the MainView, so that we can toggle the showSearchResults variable easily. That component will not re-render and we avoid the multiple calls.
> 
> This goes in pair to what I've been suggesting for a while: to have the search system accessible by a hook, and independent from the components that use it (SearchAreas and such), but that is to be done in a separate issue

Perhaps it would be best to move the auto-submit feature to another issue (since this one is just "param binding") which ideally would be dealt with after your proposed change to the search system.

imnotteixeira:
> > > I came up with the solution of using a global boolean to check weather we were on the first ever render or not (relying on mounting alone, like with `useEffect(..., [])`, does not work because apparently the component gets re-mounted after a submit, which would lead to infinite requests) but this seems like black magic. However, i have not been able to find a reasonable solution to this: any ideas ?
> > 
> > Yes! Welcome to React :)
> > 
> > That's a nice feature that hints you to not put this behavior in that component: why simulate a form submission if the user hasn't submitted anything? We can still call our own "redux-linked" search function elsewhere (I'd say on HomePage level, next to the MainView, so that we can toggle the showSearchResults variable easily. That component will not re-render and we avoid the multiple calls.
> > 
> > This goes in pair to what I've been suggesting for a while: to have the search system accessible by a hook, and independent from the components that use it (SearchAreas and such), but that is to be done in a separate issue
> 
> Perhaps it would be best to move the auto-submit feature to another issue (since this one is just "param binding") which ideally would be dealt with after your proposed change to the search system. 

Honestly I don't see the advantage of having one without the other. It's not worth it to merge this without having the auto search. The params are useless otherwise

Naapperas:
There is a bug where sometimes the scroll fails or the page snaps back to the top after the scroll occurs: I suspect this is because of the offer list being mounted but I do not know how to fix this, gotta look further into it

Naapperas:
Change the URL copy "flow" to this:

Naapperas:
I'll have to make a video demonstrating the features developed unless someone can hook up a valid backend to the Netlify preview

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -35,20 +37,40 @@ export const AdvancedSearchController = ({
     resetAdvancedSearchFields, onSubmit, searchValue, setSearchValue, onMobileClose,
 }) => {
 
+    // TODO: see if it is better to change the URL only when we are submitting the form
+    // TODO: make it so the URL can be copied to the clipboard
```
Naapperas, line 41: Oops, need to address/remove these

src/components/HomePage/MainView.spec.js
```diff
@@ -24,10 +26,12 @@ describe("Main View", () => {
...
-            <MainView
-                scrollToProductDescription={scrollToProductDescription}
-                showSearchResults={showSearchResults}
-            />,
+            <MemoryRouter initialEntries={["/"]}>
+                <MainView
+                    scrollToProductDescription={scrollToProductDescription}
+                    showSearchResults={showSearchResults}
+                />
+            </MemoryRouter>,
```
BrunoRosendo, line 34: Do you mind explaining the use of `MemoryRouter`? I'm not familiar

src/components/HomePage/SearchArea/SearchArea.js
BrunoRosendo: Remove these

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -35,20 +37,40 @@ export const AdvancedSearchController = ({
...
+    const {
+        queryParams,
+        setJobType: actualSetJobType,
+        setJobDuration: actualSetJobDuration,
+        setFields: actualSetFields,
+        setShowJobDurationSlider: actualSetShowJobDurationSlider,
+        setTechs: actualSetTechs,
+        resetAdvancedSearchFields: actualResetAdvancedSearchFields,
+        setSearchValue: setUrlSearchValue,
```
BrunoRosendo, line 50: Why doesn't the simple names on the left work? I honestly didn't think too much, it's faster to ask

src/components/HomePage/SearchArea/SearchArea.js
```diff
@@ -60,12 +82,33 @@ export const AdvancedSearchController = ({
...
+
+        if (queryParams.jobType) setJobType(queryParams.jobType);
+        if (queryParams.fields) setFields(queryParams.fields);
+        if (queryParams.technologies) setTechs(queryParams.technologies);
+
+        if (queryParams.searchValue) {
+            setSearchValue(queryParams.searchValue);
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
BrunoRosendo, line 111: Is this fine?

src/components/HomePage/SearchArea/useUrlSearchParams.js
```diff
@@ -0,0 +1,127 @@
...
+            ...changes,
+        };
+
+        history.replace({
+            ...location, search: qs.stringify(newQueryParams, {
+                skipNulls: true,
+                arrayFormat: "brackets",
+            }),
+        });
+    }, 350), [history]);
```
BrunoRosendo, line 44: Maybe put the 350 in a constant along with the explanation

src/components/HomePage/SearchArea/useUrlSearchParams.js
```diff
@@ -0,0 +1,127 @@
...
+
+    const clearURLFilters = useCallback(
+        throttle(
+            (location) => history.replace({ ...location, search: "" }),
+            350
+        ), [history]);
+
+    // destructure input here so mapDispatchToProps.setJobType is 'normalized' with respect to the other dispatch functions
```
BrunoRosendo, line 52: I didn't get it, what do you mean?

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.js
```diff
@@ -0,0 +1,65 @@
...
+    const fullURL = window.location.href;
+
+    const copyToClipboard = useCallback(() => {
+        navigator.clipboard.writeText(fullURL);
+
+        dispatch(addSnackbar({
+            message: "Copied to clipboard",
+            options: {
+                autoHideDuration: 1500,
+            },
```
BrunoRosendo, line 47: Did you manage to only change the duration in this snackbar?

src/components/HomePage/URLSearchParamsParser.js
```diff
@@ -0,0 +1,46 @@
...
+const URLSearchParamsParser = ({ showSearchResults }) => {
+
+    const location = useLocation();
+
+    const offerSearch = useSelector((state) => state.offerSearch);
+    const validQueryParams = Object.keys(offerSearch);
+
+    const queryParams = useMemo(() => qs.parse(location.search, {
+        ignoreQueryPrefix: true,
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
BrunoRosendo, line 19: Same question as above

src/components/HomePage/URLSearchParamsParser.js
```diff
@@ -0,0 +1,46 @@
...
+    useEffect(() => {
+
+        // filter to make sure we only submit form data if valid search params are present
+        if (Object.keys(queryParams).filter((key) => validQueryParams.includes(key)).length) {
+            searchOffers(SearchResultsConstants.INITIAL_LIMIT);
+
+            showSearchResults();
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
BrunoRosendo, line 41: Do we still want `useEffect()` in cases like this one? I'm not very familiar but maybe @monkin77 or @imnotteixeira could give some thoughts

src/pages/HomePage.js
```diff
@@ -1,65 +1,47 @@
...
-            },
-            searchResults: {
-                scrollSnapAlign: "start",
-
-                // don't apply snap scroll to the bottom of the search
-                // results, so that it's possible to see the footer
-                scrollMarginBottom: "100vh",
-            },
-        }
-));
```
BrunoRosendo, line 27: Where did the styles go?

BrunoRosendo (changes requested):
Good job with this PR! A lot of work was put into this.

I'm not sure design wise so I'll do the review based on functionalities and code.
We really need to get someone from the image department to do these kinds of things for us.

Regarding tests, I only checked what you were testing and it looked pretty good.

src/components/HomePage/MainView.spec.js
Naapperas: We need to have a Router, no matter what type it is, around every component that uses the AdvancedSearch because of the use of `useLocation` and `useHistory`: in Jest these do not work because JSDOM does not provide values for `location` and `history`, things which are handled by a Router. Besides that, a `MemoryRouter` lets us input directly the history stack, which can be useful for the tests.

src/components/HomePage/SearchArea/SearchArea.js
Naapperas: The names on the left are the same names as the props passed in to the controller, which relate to the "normal" functions. It was just a way to differentiate the names of the "normal" and "URL-changing" functions.

src/components/HomePage/SearchArea/SearchArea.js
Naapperas: Yes, because we only want to process the URL values once. Maybe I misunderstood your question...

src/components/HomePage/SearchArea/useUrlSearchParams.js
Naapperas: In the `mapDispatchToProps` function, every function received the exact value that needed to be dispatched except for `setJobType` which received a DOM Event. I did this in order to have the function definitions "coherent" with eachother.

src/components/HomePage/URLSearchParamsParser.js
Naapperas: This as the same thought behind as the code above: we only need this to run once, so we should not have any dependencies.

src/components/HomePage/URLSearchParamsParser.js
Naapperas: We want to imperatively trigger the offer search, that needs to be wrapped in a `useEffect`.

src/pages/HomePage.js
Naapperas: These styles relate to the "scroll-snap" feature that was messing up this PR, I've removed them here because this is more prioritary but created an issue to see if we can bring that back, check #273

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.js
Naapperas: When I tested locally it worked as expected but then on another occasion it was showing some weird behavior, gotta look further into it.

BrunoRosendo:
I think most of my questions are answered, just look into the small changes and I'll approve

src/components/HomePage/SearchArea/useUrlSearchParams.js
```diff
@@ -5,6 +5,8 @@ import { useLocation, useHistory } from "react-router-dom";
 
 import { throttle } from "lodash";
 
+const HISTORY_REPLACE_THROTTLE_DELAY = 350; // milliseconds
```
BrunoRosendo, line 8:
```suggestion
const HISTORY_REPLACE_THROTTLE_DELAY_MS = 350;
```
I think we have this format in other places

BrunoRosendo (changes requested):

BrunoRosendo (approved):

PR #260, opened 2022-08-17
Feature/recover password

DoStini:


Naapperas:
Can you post some prints of the feature ?

DoStini:
> 

There you go

https://user-images.githubusercontent.com/23351316/185762573-d9545688-5880-40dd-b240-c38cb41d1526.mov

BrunoRosendo:
> > 
> 
> There you go
> Screen.Recording.2022-08-20.at.21.01.44.mov

From watching the video, the FE flow doesn't seem to match the BE implementation. I think this would make sense if the users received a simple short code. I don't think it makes sense for them to past this huge token into the field. I prefer the flow where they just click on the link, or instead simplify the token somehow (I'm not sure if that could have security concerns)

DoStini:
> 

The BE needs to be updated to follow this login. I will see if the code can be shortened

Naapperas:
I'm with Bruno, pasting the link could be cumbersome, if we somehow could just click the link and proceed with the rest of the flow we should do that.

DoStini:
> I'm with Bruno, pasting the link could be cumbersome, if we somehow could just click the link and proceed with the rest of the flow we should do that.

I'll try a different approach, when clicking on the link, redirect with modal opened on the finish part.

> Also, the transition from "new password" to "login" seems weird but that might be just me

How so? I think it makes sense to redirect to the login "page"

DoStini:
https://user-images.githubusercontent.com/23351316/187799892-4a240ed5-b663-406c-b657-bdfde7de0360.mov

src/pages/HomePage.js
```diff
@@ -41,6 +48,15 @@ export const HomePage = () => {
         document.getElementsByTagName("html")[0].style.scrollSnapType = "y proximity";
     }, [searchResultsRef, showSearchResults]);
 
+    useEffect(() => {
+        if (openPasswordRecoveryModal && token) {
+            history.replace({ pathname: "/" });
+            dispatch(toggleAuthModal(2));
```
Naapperas, line 54: This `2` can cause confusion being left as is (I had to look at the definition of the action function to understand it was a page number). Maybe change the action function name or turn this into a constant.

src/components/Navbar/Auth/PasswordRecoveryFinishForm.spec.js
```diff
@@ -0,0 +1,203 @@
...
+
+            await act(async () => {
+                await fireEvent.change(getByLabelText("New Password"), { target: { value: "password123" } });
+            });
+            await act(async () => {
+                await fireEvent.change(getByLabelText("Token"), { target: { value: "some-token" } });
+            });
+            await act(async () => {
+                await fireEvent.click(getByRole("button", { name: "Recover Password" }));
+            });
```
Naapperas, line 72: Couldn't these have been in the same `act(...)` call ?

Naapperas: I pointed it out here but there are other places where this pattern is repeated.

Naapperas (changes requested):

Naapperas:
@DoStini is this ready for review ?

src/components/Navbar/Auth/PasswordRecoveryFinishForm.spec.js
```diff
@@ -0,0 +1,195 @@
...
+
+jest.mock("../../../services/auth");
+jest.mock("react-router-dom", () => {
+    const original = jest.requireActual("react-router-dom");
+    return {
+        ...original,
+        useParams: jest.fn().mockReturnValue({ token: "test123" }),
+    };
+});
+
```
Naapperas, line 22: In this function why did you mock `useParams` ? I understand that this and other `react-router` functions are not defined in JSDom but if you are using `BrowserRouter`s to wrap your test components the mock should not be necessary, unless I'm overlooking something important.

Naapperas (approved):

PR #272, opened 2022-10-08
Analytics for when a user opens the ApplyURL link

### LINKED ISSUE #267: Analytics for when a user opens the `ApplyURL` link
Similar to #223, we should also have an event for when a user opens the one click apply URL. This information could later on be given to the company

monkin77:
Closes #267

### DIFF
--- a/src/components/HomePage/SearchResultsArea/Offer/OfferInteractionOptions.js
+++ b/src/components/HomePage/SearchResultsArea/Offer/OfferInteractionOptions.js
@@ -5,6 +5,7 @@ import { Visibility, VisibilityOff, Block, Edit } from "@material-ui/icons";
 import { Link } from "react-router-dom";
 import OfferApplyButton from "./OfferApplyButton";
 import useToggle from "../../../../hooks/useToggle";
+import { recordApplyURLVisit } from "../../../../utils/analytics";
 
 const OfferInteractionOptions = ({
     loading,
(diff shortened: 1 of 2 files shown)


Naapperas (approved):

src/components/HomePage/SearchResultsArea/Offer/OfferInteractionOptions.js
```diff
@@ -51,6 +52,7 @@ const OfferInteractionOptions = ({
     const handleApplyURLRedirect = useCallback(() => {
         setClosedRedirectDialog();
         window.open(offer.applyURL, "_blank", "noopener");
+        if (offer) recordApplyURLVisit(offer._id, offer.title, offer.ownerName);
```
BrunoRosendo, line 55: is the `if` needed?

BrunoRosendo (approved):
This looks good, assuming you tested it on the GA page (I didn't). I left a question regarding a condition

src/components/HomePage/SearchResultsArea/Offer/OfferInteractionOptions.js
monkin77: I'm not completely sure, but I added it as a safeguard since similar verifications are done across this file to the `offer` variable
image

BrunoRosendo: Well, it doesn't hurt but I think the program will crash in line 54 if the offer is null. The nullable offer should be handled in line 78 and never be null. We should maybe instead have something like `if (!offer) return;` on the top of the function

Naapperas: @monkin77 address this so we can merge this PR.

BrunoRosendo (approved):

PR #274, opened 2022-10-17
250 unable to remove optional offer fields

AvilaAndre:
Vacancies are done, the default value is 0, but if it has no value it is not shown anyway, whenever the offer is edited again the value is set to null. Like other optional fields I'm in doubt if I should set them as null when creating (on the create form), how they are right now makes them nonexistent in the database on creation.
Also, isPaid and jobStartDate are set to null when proper input is not provided.
About the applyURL, I saw that it is optional on the database, therefore I had to modify the util that ran the regex to ignore the verification if the value was null or an empty string.
Lastly, vacancies had a visual bug where if it's value was 0 there would show a simple text 0 where the vacancies should be, but that was also fixed by checking if there are more than zero vacancies.

### DIFF
--- a/src/components/HomePage/SearchResultsArea/Offer/OfferDetails.js
+++ b/src/components/HomePage/SearchResultsArea/Offer/OfferDetails.js
@@ -214,7 +214,7 @@ const OfferDetails = ({
                     </Tooltip>
                 </Grid>
                 {
-                    !loading && (offer.vacancies > 0) &&
+                    !loading && offer.vacancies &&
                     (
                         <Grid item xs={12} md={6}>
                             <Tooltip title="Vacancies" placement="left">
--- a/src/components/HomePage/SearchResultsArea/Offer/OfferDetails.js
+++ b/src/components/HomePage/SearchResultsArea/Offer/OfferDetails.js
@@ -214,15 +214,17 @@ const OfferDetails = ({
                     </Tooltip>
                 </Grid>
                 {
-                    !loading && offer.vacancies &&
-                    <Grid item xs={12} md={6}>
-                        <Tooltip title="Vacancies" placement="left">
-                            <Typography display="inline" variant="body1" color="secondary">
-                                <FindInPage className={classes.iconStyle} />
-                                {`${offer.vacancies} ${offer.vacancies === 1 ? "vacancy" : "vacancies"}`}
-                            </Typography>
-                        </Tooltip>
-                    </Grid>
+                    !loading && (offer.vacancies > 0) &&
+                    (
+                        <Grid item xs={12} md={6}>
+                            <Tooltip title="Vacancies" placement="left">
+                                <Typography display="inline" variant="body1" color="secondary">
+                                    <FindInPage className={classes.iconStyle} />
+                                    {`${offer.vacancies} ${offer.vacancies === 1 ? "vacancy" : "vacancies"}`}
+                                </Typography>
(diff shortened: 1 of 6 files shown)


BrunoRosendo:
@AvilaAndre I haven't looked too deeply but I believe the `vacancies` field should also be set as `null` (instead of 0). 0 has a real meaning (the company actually has 0 vacancies)

AvilaAndre:
> @AvilaAndre I haven't looked too deeply but I believe the `vacancies` field should also be set as `null` (instead of 0). 0 has a real meaning (the company actually has 0 vacancies). This is also supported by the BE now.
> 
> You should also take a look at the failing tests and see what's up

With isPaid being nullable now, the test was failing because it was not expecting isPaid to exist (before being nullable its value was undefined and therefore wasn't sent through the request).

Naapperas:
@AvilaAndre is this ready for review ?

src/components/HomePage/SearchResultsArea/Offer/OfferDetails.js
```diff
@@ -214,15 +214,17 @@ const OfferDetails = ({
...
-                    !loading && offer.vacancies &&
-                    <Grid item xs={12} md={6}>
-                        <Tooltip title="Vacancies" placement="left">
-                            <Typography display="inline" variant="body1" color="secondary">
-                                <FindInPage className={classes.iconStyle} />
-                                {`${offer.vacancies} ${offer.vacancies === 1 ? "vacancy" : "vacancies"}`}
-                            </Typography>
-                        </Tooltip>
-                    </Grid>
+                    !loading && (offer.vacancies > 0) &&
```
BrunoRosendo, line 217:
```suggestion
                    !loading && offer.vacancies &&
```
I think this should be reverted

src/utils/offer/OfferUtils.js
```diff
@@ -20,5 +20,5 @@ export const validApplyURL = (val) => {
     const httpRegex = /^https?:\/\/\S+\.\S+$/;
     const emailRegex = /^mailto:(\S+@\S+)$/;
 
-    return httpRegex.test(val) || emailRegex.test(val);
+    return httpRegex.test(val) || emailRegex.test(val) || (val === "") || (val === null);
```
BrunoRosendo, line 23: Why does the `(val === "")` need to be allowed?
And what about the `(val === null)`? Is yup's validation in https://github.com/NIAEFEUP/nijobs-fe/blob/7ca25fac86ef595b7e4976ded0ffa1504c7ca18d/src/components/Offers/Edit/EditOfferSchema.js#L76 not enough?

BrunoRosendo (changes requested):

src/utils/offer/OfferUtils.js
AvilaAndre: You cannot submit the edit offer form because it expects the value to start with _https_ or _mailto:_, therefore, in order to make it optional it should accept the "" value. This happens when you edit the offer the first time, the second time, it fetches the value which is null, so it should accept the null value too, in case the user submits without changing the value.

Naapperas: Have you looked further into this ?

Naapperas (approved):
LGTM. @BrunoRosendo please take a look at the changes made.

BrunoRosendo (approved):
good job

PR #276, opened 2022-11-12
Query filtering hotfix

Naapperas:
Corrected the query parameters' filtering. Previously it allowed to trigger an auto-submit using internal offerSearch fields (like `error` or `offers`)


BrunoRosendo:
@Naapperas Is it possible to write some kind of test for this? So we guarantee it never happens again

Naapperas:
> @Naapperas Is it possible to write some kind of test for this? So we guarantee it never happens again

Yes it is. I already wrote some tests to verify that the auto-submit feature worked as expected só its just a matéria of extensing those.

src/components/HomePage/URLSearchParamsParser.js
```diff
@@ -7,12 +7,21 @@ import { useSelector } from "react-redux";
...
 
     const offerSearch = useSelector((state) => state.offerSearch);
-    const validQueryParams = Object.keys(offerSearch);
+
+    const validQueryParams = Object.keys(offerSearch).filter((field) => !(invalidParams.includes(field)));
+    if (validQueryParams.includes("jobDuration")) {
+        validQueryParams.splice(validQueryParams.indexOf("jobDuration"), 1);
+        validQueryParams.push("jobMaxDuration");
+        validQueryParams.push("jobMinDuration");
+    }
```
BrunoRosendo, line 24:
```suggestion
    if (validQueryParams.includes("jobDuration")) {
        validQueryParams.splice(validQueryParams.indexOf("jobDuration"), 1, "jobMaxDuration", "jobMinDuration");
    }
```
This should be the same. What do you think?

BrunoRosendo (changes requested):
I left a small comment and requested some tests

BrunoRosendo (approved):

PR #284, opened 2022-12-05
Fixed incorrect offeSearch state after returning to Home Page

Naapperas:

PR #286, opened 2022-12-06
Feature/239 evident see offers wo user parameters

### LINKED ISSUE #239: Make more evident how to see offers without search parameters
Sometimes, the users of NIJobs just want to see the offers available on the website, but don't have any offer in mind, so they don't want to enter search parameters.

We should make it more evident that the user can still see offers when clicking the search button without introducing any search parameters. After #172, this will be like a feed of offers.

We can, for example, add some information to the hint text of the search bar. I'd like to see some suggestions on this.

tomaspalma:
Closes #239.  The solution implemented was displaying a "search" button when there were no user specified filters nor any value in the search bar. Then, when the user chooses a filter or writes something in the search bar, it changes the button text to "search"

image

image

image

### DIFF
--- a/src/components/HomePage/SearchArea/SearchArea.spec.js
+++ b/src/components/HomePage/SearchArea/SearchArea.spec.js
@@ -10,10 +10,12 @@ import {
 } from "../../../actions/searchOffersActions";
 import { createTheme } from "@material-ui/core";
 import { renderWithStoreAndTheme, screen, fireEvent, act } from "../../../test-utils";
-
 import { MemoryRouter } from "react-router-dom";
 
+import PropTypes from "prop-types";
+
 import qs from "qs";
+import { INITIAL_JOB_TYPE } from "../../../reducers/searchOffersReducer";
 
 // eslint-disable-next-line react/prop-types
 const RouteWrappedContent = ({ children, url = "/" }) => (
@@ -21,6 +23,47 @@ const RouteWrappedContent = ({ children, url = "/" }) => (
         {children}
     </MemoryRouter>
 );
+
+const SearchAreaWrapper = ({
+    searchValue = "", jobType = INITIAL_JOB_TYPE, jobDuration = [null, null], filterJobDuration = false,
+    showJobDurationSlider = false, fields = [], technologies = [], setShowJobDurationSlider = () => { },
+    setTechs = () => { }, setJobDuration = () => { }, setFields = () => { }, setJobType = () => { },
+    setSearchValue = () => { }, onSubmit = () => {},
+}) => (
+    <SearchArea
+        searchValue={searchValue}
+        jobType={jobType}
+        jobDuration={jobDuration}
+        filterJobDuration={filterJobDuration}
+        fields={fields}
+        technologies={technologies}
+        showJobDurationSlider={showJobDurationSlider}
+        setShowJobDurationSlider={setShowJobDurationSlider}
+        setTechs={setTechs}
+        setJobDuration={setJobDuration}
+        setFields={setFields}
+        setJobType={setJobType}
+        setSearchValue={setSearchValue}
+        onSubmit={onSubmit}
+    />
+);
+
+SearchAreaWrapper.propTypes = {
+    onSubmit: PropTypes.func,
+    searchValue: PropTypes.string.isRequired,
+    jobType: PropTypes.string,
+    setSearchValue: PropTypes.func.isRequired,
+    setJobDuration: PropTypes.func.isRequired,
+    setJobType: PropTypes.func.isRequired,
+    fields: PropTypes.array.isRequired,
+    technologies: PropTypes.array.isRequired,
+    showJobDurationSlider: PropTypes.bool.isRequired,
+    setFields: PropTypes.func.isRequired,
+    setTechs: PropTypes.func.isRequired,
+    setShowJobDurationSlider: PropTypes.func.isRequired,
+    jobDuration: PropTypes.number,
+    filterJobDuration: PropTypes.bool,
+};
 
 describe("SearchArea", () => {
     let onSubmit;
@@ -34,16 +77,8 @@ describe("SearchArea", () => {
         it("should render a Paper, a Form, a Search Bar, a Search Button and Advanced Options Button", () => {
             renderWithStoreAndTheme(
                 <RouteWrappedContent>
-                    <SearchArea
+                    <SearchAreaWrapper
                         onSubmit={onSubmit}
-                        fields={[]}
-                        technologies={[]}
-                        setShowJobDurationSlider={() => { }}
-                        setTechs={() => { }}
-                        setJobDuration={() => { }}
-                        setFields={() => { }}
-                        setJobType={() => { }}
-                        setSearchValue={() => { }}
                     />
                 </RouteWrappedContent>,
                 { initialState, theme }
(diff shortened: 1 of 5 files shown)


Naapperas:
@tomaspalma You need to fix the failing test, since the current behavior will change.

tomaspalma:
> @tomaspalma You need to fix the failing test, since the current behavior will change.

It is now fixed. 

Should we also add tests to see if the searchHasUserInput variable is set to the correct value when the search bar isn't empty and when any filters are activated? because this test only tests if the behaviour is correct when a certain value of searchHasUserInput is set but it doesn't validate that the value of the searchHasUserInput is correctly set under the supposed circumstances.

Also, I have a question in the file I worked on. Why is that the component name in the SubmitSearchButton.js file is called "ShowAdvancedOptionsButton"?

Naapperas:
@tomaspalma the only thing missing in this is to write the test that we talked about in a meeting, or is there something else?

Other than that, you'll need to rebase this with develop

src/components/HomePage/SearchArea/SubmitSearchButton.spec.js
```diff
@@ -11,8 +11,16 @@ describe("SubmitSearchButton", () => {
...
+        });
+        it("should render 'span' with text = 'Show All' when searchHasInput = false", () => {
+            const button = shallow(
+                <SubmitSearchButton
+                    searchHasUserInput={false}
+                />
+            );
+            expect(button.find("span").first().text()).toEqual("Show All");
         });
     });
```
Naapperas: These tests work but if you could rewrite them using the functions from React Testing Library (i believe they are the `render` and `renderWith*` variants) it would be great.

`shallow` is a function given by the `enzyme` module which is no longer maintained and tests using methods from Enzyme should migrate to other tools: in our case that is RTL. @imnotteixeira had some work migrating most of the tests from Enzyme to RTL so if there are any questions you can ask him but I believe this is an easy fix.

Naapperas (changes requested):

src/components/HomePage/SearchArea/SubmitSearchButton.spec.js
imnotteixeira: > if you could rewrite them using the functions from React Testing Library it would be great.

I'd go further: we should have no new code with enzyme tests whatsoever. We had a lot of work migrating existing tests to RTL, and we should be using it, and trying to migrate the last few so we can finally remove the dependency on Enzyme.

Please do not merge PRs with Enzyme tests

Naapperas: Should we also change the `config/test.js` file created by the CRA `eject` command? The Enzyme globals are defined there.

imnotteixeira: Once enzyme is not used, yes

Naapperas:
@tomaspalma The failing tests may need more props that are mandatory.

Naapperas (changes requested):
LGTM, the only thing left is to fix the tests but I have explained what should be done for that so it should not be too complicated.

tomaspalma:
> @tomaspalma The failing tests may need more props that are mandatory.

I was unsure if there was something really missing on the props or whether it was some other issue because locally running npm test the result was an all-green suite. It was only failing with the ci tests. 

But now seeing what you said and looking at the other tests that render searchArea in searchArea.spec.js, the component is being rendered in those tests with some props I'm not passing in the tests I created, including the one that is saying not to be defined in the error report of the ci tests, the other props probably aren't showing as undefined because setShowJobDurationSlider appears first in the searchArea component code, so it may be that.

Update: The tests we failling locally because I had an outdated version of the program code, because I had forgotten to rebase with develop.

tomaspalma:
It is not ready yet, there's still an issue that came up after updating the changes on develop and the searchValue comes initially as undefined instead of empty string, so the show all button is not currently showing at start.

Naapperas:
> It is not ready yet, there's still an issue that came up after updating the changes on develop and the searchValue comes initially as undefined instead of empty string, so the show all button is not currently showing at start.

Maybe you can change the initial value of the redux state for the `offerSearch`, it is defined in the offer's reducer if I'm not mistaken.

Another option is to check if it is undefined.

src/components/HomePage/SearchArea/SearchArea.spec.js
```diff
@@ -55,6 +55,128 @@ describe("SearchArea", () => {
...
+                        filterJobDuration={false}
+                        offers={[]}
+                        fields={[]}
+                        technologies={[]}
+                        setShowJobDurationSlider={() => { }}
+                        setTechs={() => { }}
+                        setJobDuration={() => { }}
+                        setFields={() => { }}
+                        setJobType={() => { }}
+                        setSearchValue={() => { }}
```
Naapperas: Perhaps one thing you could do is create a separate test component that passes default props to SearchArea, kinda like this:
```js
const SearchAreaWithProps = (props = {<here you put all the props with default values or something like that>}) => <SearchArea {...props} />;
```

Then you would just need to overwrite the prop(s) that you wanted to test.

I do not recall off the top of my head if this is syntatically correct but the idea is there.

src/components/HomePage/SearchArea/SearchArea.spec.js
```diff
@@ -55,6 +55,128 @@ describe("SearchArea", () => {
...
+                        searchValue={"somevalue"}
+                        fields={[]}
+                        technologies={[]}
+                        setShowJobDurationSlider={() => { }}
+                        setTechs={() => { }}
+                        setJobDuration={() => { }}
+                        setFields={() => { }}
+                        setJobType={() => { }}
+                        setSearchValue={() => { }}
+                    />
```
Naapperas: Same thing as above.

src/components/HomePage/SearchArea/SearchArea.spec.js
```diff
@@ -55,6 +55,128 @@ describe("SearchArea", () => {
...
+                    <SearchArea
+                        fields={["field1", "field2"]}
+                        technologies={[]}
+                        setShowJobDurationSlider={() => { }}
+                        setTechs={() => { }}
+                        setJobDuration={() => { }}
+                        setFields={() => { }}
+                        setJobType={() => { }}
+                        setSearchValue={() => { }}
+                    />
```
Naapperas: Same thing as above.

src/components/HomePage/SearchArea/SearchArea.spec.js
```diff
@@ -55,6 +55,128 @@ describe("SearchArea", () => {
...
+                    <SearchArea
+                        fields={[]}
+                        technologies={["tech1", "tech2"]}
+                        setShowJobDurationSlider={() => { }}
+                        setTechs={() => { }}
+                        setJobDuration={() => { }}
+                        setFields={() => { }}
+                        setJobType={() => { }}
+                        setSearchValue={() => { }}
+                    />
```
Naapperas: Same thing as above.

src/components/HomePage/SearchArea/SearchArea.spec.js
```diff
@@ -55,6 +55,128 @@ describe("SearchArea", () => {
...
+                        fields={[]}
+                        technologies={[]}
+                        jobType={"JOB"}
+                        setShowJobDurationSlider={() => { }}
+                        setTechs={() => { }}
+                        setJobDuration={() => { }}
+                        setFields={() => { }}
+                        setJobType={() => { }}
+                        setSearchValue={() => { }}
+                    />
```
Naapperas: Same thing as above.

src/components/HomePage/SearchArea/SearchArea.spec.js
```diff
@@ -55,6 +55,128 @@ describe("SearchArea", () => {
...
+                        fields={[]}
+                        technologies={[]}
+                        showJobDurationSlider={true}
+                        setShowJobDurationSlider={() => { }}
+                        setTechs={() => { }}
+                        setJobDuration={() => { }}
+                        setFields={() => { }}
+                        setJobType={() => { }}
+                        setSearchValue={() => { }}
+                    />
```
Naapperas: Same thing as above.

Naapperas (changes requested):
I believe I have done something similar to what I have described in the review comments but perhaps I'm mistaken.

Edit: FOUND IT!! It was in this PR: https://github.com/NIAEFEUP/nijobs-fe/pull/284 and on this file `src/components/HomePage/SearchArea/AdvancedSearch/AdvancedSearchDesktop.spec.js`.

tomaspalma:
I think it's pratically done now, already fixed show all not showing at start and also added the search area wrapper to improve readiblity in the tests. Contrarily to the example you gave me, I had to specify the prop types because the linter was giving me warnings about not specifying the types for them.

Naapperas:
> I think it's pratically done now, already fixed show all not showing at start and also added the search area wrapper to improve readiblity in the tests. Contrarily to the example you gave me, I had to specify the prop types because the linter was giving me warnings about not specifying the types for them.

Just as a future reference you can instruct the linter to ignore those warnings, which is useful for cases like this one.

Naapperas (approved):
LGTM :rocket:

PR #287, opened 2022-12-07
Add README badge for code coverage

### LINKED ISSUE #264: Add README badge for code coverage
NIJobs has a pretty good test coverage of all its features but we're not showcasing it to everyone. It'd be nice to have a badge showing how much coverage we have.

One could look into [Codecov](https://about.codecov.io/) or a similar tool. We should be able to easily integrate it with our Github actions. CI. You can look at multiple projects using this, but here's an example: https://github.com/codecov/example-javascript.

It's also possible to integrate this tool with pull requests and it'll tell us how the coverage has increased or decreased. We can consider using this if we think is worth it.

ttoino:
Closes #264.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,9 @@ jobs:
     - name: Test
       run: CI=true npm run test-ci
 
+    - name: Upload code coverage to Codecov
+      uses: codecov/codecov-action@v3
+
   audit:
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


ttoino:
Code coverage reports can now be found [here](https://app.codecov.io/gh/NIAEFEUP/nijobs-fe) (they will only show up for this pull request until it is merged).

Anyone from the organization should be able to login to Codecov with Github and access all the repository settings. I've left them all as the default, but we should discuss if we want any of them changed. [Here](https://docs.codecov.com/docs/common-recipe-list) are some common configurations and [here](https://docs.codecov.com/docs/codecovyml-reference) is the config file reference.

Naapperas (approved):
LGTM

PR #302, opened 2023-02-04
Prepend mailto: when company inserts an email in one click application

### LINKED ISSUE #256: Prepend `mailto:` when company inserts an email in one click application
Currently, it's possible for a company to use an email for the one click application, but they have to write it as `malito:email@gmail.com`, which is not great.
The idea is to parse the application URL, and if it is an email, preprend `mailto:` to it before sending to the BE. Be careful to send the hint message in the offer form to reflect these changes.
For more context, check [this comment](https://github.com/NIAEFEUP/nijobs-fe/pull/249#issuecomment-1203189248)

Process-ing:
Closes #256

Screenshot from 2023-02-03 13-06-15
Screenshot from 2023-02-03 19-36-23
Screenshot from 2023-02-03 19-36-33

### DIFF
--- a/src/components/Offers/Edit/EditOfferForm.js
+++ b/src/components/Offers/Edit/EditOfferForm.js
@@ -1,5 +1,5 @@
 import React, { useCallback, useContext, useEffect, useState } from "react";
-import { OfferConstants, parseRequestErrors } from "../Form/OfferUtils";
+import { OfferConstants, parseApplyURL, parseRequestErrors } from "../Form/OfferUtils";
 import OfferForm from "../Form/form-components/OfferForm";
 import { editOffer } from "../../../services/offerService";
 import { Redirect, useLocation, useParams } from "react-router-dom";
@@ -8,7 +8,6 @@ import useOfferForm from "../../../hooks/useOfferForm";
 import { INITIAL_JOB_DURATION } from "../../../reducers/searchOffersReducer";
 import useSession from "../../../hooks/useSession";
 import EditOfferSchema from "./EditOfferSchema";
-import { MailRegex } from "../../../utils/offer/OfferUtils";
 
 export const EditOfferControllerContext = React.createContext();
 
@@ -43,17 +42,9 @@ const parseOfferForm = ({
     vacancies: vacancies || "",
     description,
     descriptionText: parseDescription(description),
-    applyURL: /^mailto:/.test(applyURL) ? applyURL.substring(7) : applyURL,
+    applyURL: applyURL.startsWith("mailto:") ? applyURL.substring(7) : applyURL,
     ...offer,
 });
-
-const parseApplyURL = (applyURL) => {
-    if (!applyURL)
-        return null;
-    if (MailRegex.test(applyURL) && /^(?!mailto:)/.test(applyURL))
-        return `mailto:${applyURL}`;
-    return applyURL;
-};
 
 export const EditOfferController = () => {
     const { id } = useParams();
--- a/src/components/Offers/Edit/EditOfferForm.js
+++ b/src/components/Offers/Edit/EditOfferForm.js
@@ -49,11 +49,11 @@ const parseOfferForm = ({
 
 const parseApplyURL = (applyURL) => {
     if (!applyURL)
-        return null
+        return null;
     if (MailRegex.test(applyURL) && /^(?!mailto:)/.test(applyURL))
-        return 'mailto:' + applyURL
-    return applyURL
-}
+        return `mailto:${applyURL}`;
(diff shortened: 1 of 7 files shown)


src/components/Offers/Edit/EditOfferForm.js
```diff
@@ -41,9 +43,18 @@ const parseOfferForm = ({
...
     description,
     descriptionText: parseDescription(description),
+    applyURL: /^mailto:/.test(applyURL) ? applyURL.substring(7) : applyURL,
     ...offer,
 });
 
+const parseApplyURL = (applyURL) => {
+    if (!applyURL)
+        return null;
+    if (MailRegex.test(applyURL) && /^(?!mailto:)/.test(applyURL))
```
BrunoRosendo: Just a suggestion, maybe this function would be more readable? It should give the same result without regex: https://www.w3schools.com/jsref/jsref_startswith.asp

What do you think?

src/components/Offers/Edit/EditOfferForm.js
```diff
@@ -41,9 +43,18 @@ const parseOfferForm = ({
     vacancies: vacancies || "",
     description,
     descriptionText: parseDescription(description),
+    applyURL: /^mailto:/.test(applyURL) ? applyURL.substring(7) : applyURL,
```
BrunoRosendo: Why do we need to remove `mailto:` when we check if it's in the string in `pareApplyURL`? Maybe I missed something, I'm not sure

src/components/Offers/New/CreateOfferForm.js
```diff
@@ -5,6 +5,15 @@ import { newOffer } from "../../../services/offerService";
...
 import CreateOfferSchema from "./CreateOfferSchema";
+import { MailRegex } from "../../../utils/offer/OfferUtils";
+
+const parseApplyURL = (applyURL) => {
+    if (!applyURL)
+        return null;
+    if (MailRegex.test(applyURL) && /^(?!mailto:)/.test(applyURL))
+        return `mailto:${applyURL}`;
+    return applyURL;
+};
```
BrunoRosendo: this is duplicated, try to move it to somewhere else, maybe OfferUtils?

BrunoRosendo (changes requested):
Looks good, I only have some minor cosmetic suggestions

src/components/Offers/Edit/EditOfferForm.js
Process-ing: Yes, it would make the code way more clean.

src/components/Offers/New/CreateOfferForm.js
Process-ing: Guess it makes sense.

src/components/Offers/Edit/EditOfferForm.js
Process-ing: It's not in `parseApplyURL`, it's to ignore `mailto:` when we edit an existing offer. I think if a company that didn't use `mailto:`in the one click application's email, it would be weird to suddenly appear when editing the offer in the future. But yes, it's not mandatory.
Screenshot from 2023-02-03 13-06-15

Naapperas: This is fine but is something that is very rarely going to happen: companies do not need and should not need to know that `mailto:` is a thing, they should just submit their email and it "magically" works. Nevertheless this already catches those rare few that could in fact write that.

Process-ing: I think you're talking about the `parseApplyURL` function. This is while loading the offer in the editing page. When the email is fetched, it comes with `mailto:` prepended, and the line mentioned removes `mailto:`.

Naapperas: That's the thing: `mailto:` should never be fetched, since its a browser thing. It should just be prepended on the client and not stored on the database, in my opinion.

However i might be missing something and it actually would make sense to store that in the DB, I'm open to ideas.

Naapperas: I just checked the issue description and it talks about sending the `mailto:` to the database. @BrunoRosendo any reason why?

BrunoRosendo: As we discussed in person, mailto is kept on the database because the value could either be en email or a URL. I understand we don't want to show it on the form, so I think this implementation is fine

BrunoRosendo (approved):
Good job! Looks good

Naapperas (approved):

Naapperas:
@Process-ing rebase

PR #309, opened 2023-02-11
Validate company application page

FranciscoCardoso913:
### Important:

This is a backend issue! See the original issue and the changes made in the backend here: https://github.com/NIAEFEUP/nijobs-be/pull/306

### Discussed solution:
    
    -When created, applications should have a state UNVERIFIED by default instead of the current PENDING state;

    -After creating an application an email with a link should be sent to the company so they can validate their application, the link should expire in 10 minutes;

    -A company should not be able to create an application with an email associated with an UNVERIFIED application whose link is yet to expire;

    -If the link of an application expires and the company tries to create another application the previous application of that company should be deleted;

    -The link should send the user to a page where the application is validated, when validating the application its state should change to pending and an email should be sent to the company.

    

### Front-End changes:

    -You can now search for unverified applications;
Screenshot from 2023-02-11 20-18-26

    -The message which appeared after creating an application has now changed to indicate the company they have to validate their application;
Screenshot from 2023-02-11 20-14-37

    -New error message when trying to create a new application with an email associated with an UNVERIFIED application whose link is yet to expire;
Screenshot from 2023-02-11 20-34-35

    -Created a new page for validating the company, the page displays various messages indicating if the validation was successful or if an error has occurred;

### States of the new page:

    -Page when loading:

Screenshot from 2023-02-11 20-22-53

    -Page when everything goes well:

Screenshot from 2023-02-11 20-25-06

     -Page when the application was already validated:

Screenshot from 2023-02-11 20-30-44

    -Page when the application does not exist(the token in the link is incorret):

Screenshot from 2023-02-11 20-27-41

    -Page when the link has expired:

Screenshot from 2023-02-11 20-26-34

    -Page when an unexpected error occurs:

Screenshot from 2023-02-11 20-28-51

### What is yet to be done:

     -Tests;

     -Google analytics;
    

### Doubts:

    -The error messages used have a style made by us, however, there already exists a template style to display errors ( a component called ErrorComponent), should we use that or keep it as it is?

    -Should we change any of the designs of the validation page?

    -Should we change any of the texts?	

    -Is everything in the correct place or is there anything that is more adequate somewhere else?

### DIFF
--- a/src/AppRouter.js
+++ b/src/AppRouter.js
@@ -139,7 +139,7 @@ const AppRouter = () => (
             <Route exact path="/apply/validate/:token/confirm" key="/apply/validate/:token/confirm">
                 <PageLayout
                     key="/apply/validate/:token/confirm"
-                    pageTitle="Validate"
+                    pageTitle="Validate Company Application"
                     layout={LayoutType.DESKTOP}
                 >
                     <ValidationPage />
--- a/src/AppRouter.js
+++ b/src/AppRouter.js
@@ -38,6 +38,7 @@ import EditOfferPage from "./pages/EditOfferPage";
 import PrivacyPolicyPage from "./pages/PrivacyPolicyPage";
 import TermsAndConditionsPage from "./pages/TermsAndConditionsPage";
 import ChangeLogPage from "./pages/ChangeLogPage";
+import ValidationPage from "./pages/ValidationPage";
 
 /**
  *
@@ -133,6 +134,15 @@ const AppRouter = () => (
                     layout={LayoutType.DESKTOP}
                 >
                     <RulesPage />
+                </PageLayout>
+            </Route>
+            <Route exact path="/apply/validate/:token/confirm" key="/apply/validate/:token/confirm">
+                <PageLayout
+                    key="/apply/validate/:token/confirm"
+                    pageTitle="Validate"
+                    layout={LayoutType.DESKTOP}
+                >
+                    <ValidationPage />
                 </PageLayout>
             </Route>
             <ProtectedRoute
--- a/src/AppRouter.js
+++ b/src/AppRouter.js
@@ -136,9 +136,9 @@ const AppRouter = () => (
                     <RulesPage />
                 </PageLayout>
             </Route>
-            <Route exact path="/apply/validate/:token/confirm" key="/apply/validate/:token/confirm">
+            <Route exact path="/apply/:token/validate" key="/apply/:token/validate">
                 <PageLayout
-                    key="/apply/validate/:token/confirm"
+                    key="/apply/:token/validate"
                     pageTitle="Validate Company Application"
                     layout={LayoutType.DESKTOP}
                 >
--- a/src/components/Apply/Company/CompanyApplicationUtils.js
+++ b/src/components/Apply/Company/CompanyApplicationUtils.js
@@ -20,7 +20,11 @@ const HumanReadableErrors = Object.freeze({
     "company-application-recently-created": "There is an application created less than 10 minutes ago associated with this email.",
 });
 
-const ValidationErrors = Object.freeze({
+const ValidationMessages = Object.freeze({
+    "success": {
+        title: "Your application has been validated successfully!",
+        text: "You should receive a confirmation email shortly. If not, please contact us at ",
+    },
     "invalid-token": {
         title: "Error! Application does not exist!",
         text: "An error has occurred while validating your application! The application you are trying to validate does not exist.",
@@ -41,8 +45,8 @@ export const getValidationError = (error) => {
     if (!error) {
         return errorMsg;
     }
-    if (typeof ValidationErrors[error] === "object") {
-        return ValidationErrors[error];
+    if (typeof ValidationMessages[description] === "object") {
+        return ValidationMessages[description];
     }
     return errorMsg;
 };
--- a/src/components/Apply/Company/CompanyApplicationUtils.js
+++ b/src/components/Apply/Company/CompanyApplicationUtils.js
@@ -23,21 +23,21 @@ const HumanReadableErrors = Object.freeze({
 const ValidationErrors = Object.freeze({
     "invalid-token": {
         title: "Error! Application does not exist!",
-        text: "An error has occured while validating your application! The application you are trying to validate does not exist,",
+        text: "An error has occurred while validating your application! The application you are trying to validate does not exist.",
     },
     "expired-token": {
         title: "Error! Link has expired!",
-        text: "An error has occured while validating your application! The link sent to you has expired, \
-        you now need to create a new application,",
+        text: "An error has occurred while validating your application. The link that was sent to you has expired." +
+            " You will need to create a new application.",
     },
     "application-already-validated": {
         title: "Application is already validated!",
-        text: "This application is already validated",
+        text: "This application is already validated. ",
     },
 });
 
 export const getValidationError = (error) => {
-    const errorMsg = { title: "Unexpected Error!", text: "An unexpected error has occured while validating your application, " };
+    const errorMsg = { title: "Unexpected Error!", text: "An unexpected error has occurred while validating your application. " };
     if (!error) {
         return errorMsg;
     }
--- a/src/components/Apply/Company/CompanyApplicationUtils.js
+++ b/src/components/Apply/Company/CompanyApplicationUtils.js
@@ -1,7 +1,5 @@
 import { validationRulesGenerator, generalHumanError } from "../../../utils";
 import { AuthConstants } from "../../Navbar/Auth/AuthUtils";
-
-
 export const CompanyApplicationConstants = {
     password: AuthConstants.password,
     motivation: {
@@ -19,6 +17,34 @@ export const generateValidationRule = validationRulesGenerator(CompanyApplicatio
 const HumanReadableErrors = Object.freeze({
     "email-already-exists": "The provided email is already associated to our platform.",
     "company-application-duplicate-email": "There is already an application associated with that email.",
+    "company-application-recently-created": "There is an application created less than 10 minutes ago associated with this email.",
 });
+
+const ValidationErrors = Object.freeze({
+    "invalid-token": {
+        title: "Error! Application does not exist!",
+        text: "An error has occured while validating your application! The application you are trying to validate does not exist,",
+    },
+    "expired-token": {
+        title: "Error! Link has expired!",
+        text: "An error has occured while validating your application! The link sent to you has expired, \
+        you now need to create a new application,",
+    },
+    "application-already-validated": {
+        title: "Application is already validated!",
+        text: "This application is already validated",
+    },
+});
+
+export const getValidationError = (error) => {
+    const errorMsg = { title: "Unexpected Error!", text: "An unexpected error has occured while validating your application, " };
+    if (!error) {
+        return errorMsg;
+    }
+    if (typeof ValidationErrors[error] === "object") {
+        return ValidationErrors[error];
+    }
+    return errorMsg;
+};
 
 export const getHumanError = (error) => generalHumanError(error, HumanReadableErrors);
--- a/src/components/Apply/Company/CompanyApplicationUtils.js
+++ b/src/components/Apply/Company/CompanyApplicationUtils.js
@@ -40,9 +40,9 @@ const ValidationMessages = Object.freeze({
     },
 });
 
-export const getValidationError = (error) => {
+export const getValidationMessage = (description) => {
     const errorMsg = { title: "Unexpected Error!", text: "An unexpected error has occurred while validating your application. " };
-    if (!error) {
+    if (!description) {
         return errorMsg;
     }
     if (typeof ValidationMessages[description] === "object") {
--- a/src/components/Apply/Company/CompanyApplicationUtils.js
+++ b/src/components/Apply/Company/CompanyApplicationUtils.js
@@ -35,8 +35,12 @@ const ValidationMessages = Object.freeze({
             " You will need to create a new application.",
     },
     "application-already-validated": {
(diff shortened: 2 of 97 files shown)


src/components/Apply/Company/CompanyApplicationUtils.js
```diff
@@ -19,6 +17,34 @@ export const generateValidationRule = validationRulesGenerator(CompanyApplicatio
...
+const ValidationErrors = Object.freeze({
+    "invalid-token": {
+        title: "Error! Application does not exist!",
+        text: "An error has occured while validating your application! The application you are trying to validate does not exist,",
+    },
+    "expired-token": {
+        title: "Error! Link has expired!",
+        text: "An error has occured while validating your application! The link sent to you has expired, \
+        you now need to create a new application,",
+    },
```
DoStini: Now that I'm thinking about this, an expired token should be considered as a NOT FOUND application. This is more backend wise

DoStini: Don't use a title, put these in the HumanReadableErrors

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+        textAlign: "center",
+
+    },
+    button: {
+        background: "rgb(250,70,70)",
+        color: "white",
+        marginTop: "3vh",
+
+    },
+}));
```
DoStini: This looks like way too much CSS, we will have to analyse this to see what can be replace with Material UI defaults

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+const useStyles = (isMobile) => makeStyles((theme) => ({
+    content: {
+        padding: isMobile ? theme.spacing(2, 2) : theme.spacing(3, 3),
+        display: "flex",
+        flexDirection: "column",
+        alignItems: "center",
+        justifyContent: "center",
+    },
+    title: {
+        color: "rgb(80,80,80)",
```
DoStini: Use theme colors

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+        fontSize: "23px",
```
DoStini: Hard coded font size, simply use `<Typography>` with `variant` as `h1,h2,...`

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+    text: {
+        color: "rgb(100,100,100)",
```
DoStini: hardcoded color

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+        fontSize: "15px",
```
DoStini: hardcoded font size -> non responsive

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+        maxWidth: isMobile ? "100%" : "80%",
+        textAlign: "center",
+
+    },
+    button: {
+        background: "rgb(250,70,70)",
+        color: "white",
```
DoStini: same as colors

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+                setLoading(false);
+                setSuccess(true);
+                await validateApplication(token);
+            } catch (error_) {
+                setError(error_[0].msg);
+                setLoading(false);
+                setSuccess(false);
+            }
+
+        }
```
DoStini: Why are you creating a method and calling it immediately after?

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+    const [error, setError] = useState("");
+
+
+    useEffect(() => {
+        async function validate() {
+            try {
+                setLoading(false);
+                setSuccess(true);
+                await validateApplication(token);
+            } catch (error_) {
```
DoStini:
```suggestion
            } catch (err) {
```

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+                <CircularProgress  />
+            </CardContent>
+        );
+    } else if (success) {
+            return  (
+                <CardContent className={classes.content}>
+                    <h2 className={classes.title}>Your application has been validated successfully! </h2>
+                    <span className={classes.text}>
+                        You should receive a confirmation email shortly. If not, please contact us:
+                         <a href="mailto:nijobs@aefeup.pt"> nijobs@aefeup.pt</a>
```
DoStini: I Think there's a constant for this somewhere, please check it

DoStini: Use `<Link>` component with href

src/pages/ValidationPage.js
```diff
@@ -0,0 +1,113 @@
...
+                    </span>
+                    <Button className={classes.button} variant="contained" onClick={() => setButtonPressed(true) }>
+                        Click here to go to Home page
+                    </Button>
```
DoStini: Use a `<Link>` component instead of state and redirect. Use with RouterLink component. Check other examples

src/components/Review/Applications/ApplicationsReviewTableSchema.js
```diff
@@ -6,7 +6,9 @@ export const columns = {
 };
 
 export const ApplicationStateLabel = Object.freeze({
+    UNVERIFIED: "Unverified",
```
DoStini, line 9: Do we want Unverified aaplications to show uo in the admin panel? I think it shouldn't

- Still not a valid application
- If we are victims of spam, they will show up

src/AppRouter.js
```diff
@@ -135,6 +136,15 @@ const AppRouter = () => (
                     <RulesPage />
                 </PageLayout>
             </Route>
+            <Route exact path="/apply/validate/:token/confirm" key="/apply/validate/:token/confirm">
+                <PageLayout
+                    key="/apply/validate/:token/confirm"
+                    pageTitle="Validate"
```
DoStini:
```suggestion
                    pageTitle="Validate Company Application"
```

DoStini (changes requested):

src/components/Review/Applications/ApplicationsReviewTableSchema.js
DoStini: cc @BrunoRosendo @Naapperas For more opinions

BrunoRosendo: What's the flow here? We can only accept them once they've confirmed their emails? Imo that's too much friction for a website where it's already hard to bring companies into. They need to register, confirm email, wait for the application, and finish registration before starting to create offers.

What if, instead of that, we let the companies use the website without confirming email (just like today), and use the `isVerified` prop like a Twitter verification? The company could have a warning (snack bar) after logging in and the students (and admins) would be able to know if the company is fully trust worthy.

FranciscoCardoso913: If i am not mistaken, in a reunion it was discussed that Unverified should be a state like PENDING APROVED and REJECTED , so you can filter your application to show unverified application or others. About the spam, as it stands only one application with the same email can exist , however that way we are still vonorable to be spamed with applications with different emails, one solution to that would be deleting all application that are unverified after 10 minutes (when you can longer validate them) , but of course that way having an unverified state would no longer make sense.

As it stands when creating an application your application will be Unverified and cannot be reviewed, you will than receive an email to verifie it, if you do so then your application become PENDING and is ready to be reviewed. 

The idea of being like twitter is an interesting ideia, however it may be better to discuss it with the rest of the team as it is a concept completely different from the original ideia( it may be better to discussed it on the next reunion). In that case isVerified should probably be on the company instead of the application and the way companies verifie their companies should be diferent and only done after they are aproved. The email with a link that expires in 10 minutes which is sent when the application is created would no longer make a lot of sense in that case, maybe we could have a button in the company profile where you could validate your company, that way we could still use the email with the link to validate it

src/pages/ValidationPage.js
FranciscoCardoso913: UseEffect can't be async, because of that if you want to have an await inside of an useEffect you have to use an auxiliar async function , .then() would also work as an alternative

src/components/Apply/Company/CompanyApplicationUtils.js
FranciscoCardoso913: Our ideia was to make it easier for the company to identify what went wrong, that way they can see if the token has expired or if the token is just wrong, otherwise it will seem as it is the token is wrong when in reality it has only expired.
Okok we will change that

src/components/Review/Applications/ApplicationsReviewTableSchema.js
BrunoRosendo: Yes, I think we should discuss this again. I know it wasn't the original idea but I was re-thinking about it myself

dsantosferreira: > What's the flow here? We can only accept them once they've confirmed their emails? Imo that's too much friction for a website where it's already hard to bring companies into. They need to register, confirm email, wait for the application, and finish registration before starting to create offers.
> 
> What if, instead of that, we let the companies use the website without confirming email (just like today), and use the `isVerified` prop like a Twitter verification? The company could have a warning (snack bar) after logging in and the students (and admins) would be able to know if the company is fully trust worthy.

I really like the Twitter verification idea but there is something that is bothering me. I'm sorry if what I'm about to say is completely wrong, but wouldn't the spam issue persist in the offers page, as each account can potentially create up to five concurrent offers? Wouldn't the offer page or even our database become way too bloated?
An idea that i had was to develop a scheduled function that would run every hour or so (not sure of the appropriate time interval) that would erase all expired applications.

src/pages/ValidationPage.js
Naapperas: Also, `makeStyles` returns a function that can optionally take arguments, which in this case would be `isMobile`, so there is no need to apply currying here: that argument is available to use if instead of an object you use a function as the style, for instance:

Instead of
```js
text: {
    ...
    maxWidth: isMobile ? "100%" : "80%",
}
```
you could have
```js
text: ({isMobile}) => ({
    ...
    maxWidth: isMobile ? "100%" : "80%",
})
```
which makes it so only the styles that may need it use it.

I'm not entirely sure about the syntax but I believe this is a thing. Correct me if I'm wrong.

This is not so much a problem as it is just advice/letting you know.

src/components/Review/Applications/ApplicationsReviewTableSchema.js
Naapperas: I like the Twitter thing, in addition to said pop-up/snackbar, perhaps we could/should restrict some actions made by unverified companies, but this may cause more harm than good. It should be discussed in the weekly meting but I personally am leaning towards what @BrunoRosendo suggested.

FranciscoCardoso913:
## Important
The workflow regarding this issue has been changed
You can check it here: https://github.com/NIAEFEUP/nijobs-be/pull/306#issuecomment-1447121249

Naapperas:
@FranciscoCardoso913 @dsantosferreira any progress on this?

dsantosferreira:
We have our first version of the warnings shown to users whose accounts aren't yet verified in the "Create an offer" and "My offers" pages. Feel free to share your opinions!

warning1
warning2

BrunoRosendo:
> We have our first version of the warnings shown to users whose accounts aren't yet verified in the "Create an offer" and "My offers" pages. Feel free to share your opinions!
> 
> warning1 warning2

I like how it turned out! I just have some doubts about the icon. Is it also from material UI? The style seems a bit off from the rest of your icons, is it imported from `@material-ui/icons` or somewhere else?

Assuming it is the right one, I guess it's the one on the right. Do you mind trying out the first one and see how it looks like?
image

dsantosferreira:
> 

The icon on the images I sent is the default one used by the alert component from material UI when the severity attribute is set to "warning". It is probably the "WarningAmber" icon, as you referred.
Here is the alert with the filled version of the icon. I prefer the outlined one but tell me what you think!

image

Naapperas:
> > 
> 
> The icon on the images I sent is the default one used by the alert component from material UI when the severity attribute is set to "warning". It is probably the "WarningAmber" icon, as you referred. Here is the alert with the filled version of the icon. I prefer the outlined one but tell me what you think!
> 
> image

Honestly, I prefer the filed version of the icon.

BrunoRosendo:
> > 
> 
> The icon on the images I sent is the default one used by the alert component from material UI when the severity attribute is set to "warning". It is probably the "WarningAmber" icon, as you referred. Here is the alert with the filled version of the icon. I prefer the outlined one but tell me what you think!
> 
> image

I do think the filled version looks more like the remaining icons we use (from what I recall, but feel free to check it out)

Naapperas (changes requested):
Remove the `console.log` statements

src/components/Apply/Company/CompanyApplicationUtils.js
```diff
@@ -19,6 +17,38 @@ export const generateValidationRule = validationRulesGenerator(CompanyApplicatio
...
+        text: "An error has occurred while validating your application! The application you are trying to validate does not exist.",
+    },
+    "expired-token": {
+        title: "Error! Link has expired!",
+        text: "An error has occurred while validating your application. The link that was sent to you has expired." +
+            " You will need to create a new application.",
+    },
+    "application-already-validated": {
+        title: "Application is already validated!",
+        text: "This application is already validated. ",
```
Naapperas:
```suggestion
        title: "Application was already validated!",
        text: "This application was already validated. ",
```

Naapperas (changes requested):

PR #313, opened 2023-03-01
Form errors should be made more explicit

### LINKED ISSUE #306: Form errors should be made more explicit
When a company tries to create an offer and they have made an error, a message related to that error appears. However, the page does not scroll and there is no clear indication about that or other errors, so companies have a hard time figuring out what to fix. That should be changed.

Process-ing:
Closes #306 

Now, when someone tries to submit a new offer or offer edit with errors, the form scrolls automatically to the corresponding field.
Screenshot from 2023-03-01 14-49-24
Screenshot from 2023-03-01 14-49-41
Screenshot from 2023-03-01 14-49-48


src/components/Offers/Form/form-components/OfferForm.js
```diff
@@ -102,12 +102,36 @@ const OfferForm = ({ context, title }) => {
...
+        if (Object.keys(requestErrors).length !== 0) {
+            const element = document.getElementById(Object.keys(requestErrors)[0]);
+            if (element?.scrollIntoView) {
+                element.scrollIntoView({ behavior: "smooth" });
+            }
+            if (element?.focus) {
+                element.focus();
+            }
+        }
+    }, [requestErrors]);
```
BrunoRosendo:
```suggestion
    useEffect(() => {
        if (Object.keys(errors).length !== 0) {
            let element = document.getElementById(Object.keys(errors)[0]);
            if (!element) {
                element = document.getElementById(Object.keys(requestErrors)[0]);
            }
            if (element?.scrollIntoView) {
                element.scrollIntoView({ behavior: "smooth" });
            }
            if (element?.focus) {
                element.focus();
            }
        }
    }, [errors, requestErrors]);
```
I think something like this is simpler, there's a lot of repeated logic here. I didn't test it so make sure the code works

src/components/Offers/Form/form-components/OfferForm.js
```diff
@@ -311,35 +337,39 @@ const OfferForm = ({ context, title }) => {
...
-                                            controllerName="contacts"
-                                            onAdd={appendContact}
-                                            onRemove={removeContact}
-                                            getValues={getValues}
-                                            control={control}
-                                            errors={errors.contacts || requestErrors.contacts}
-                                            disabled={formDisabled}
-                                            addEntryBtnTestId="contacts-selector"
-                                        />
+                                        <div id="contacts">
```
BrunoRosendo: What are the new divs (contacts and requirements) used for? i didn't quite get it

BrunoRosendo (changes requested):
Great job so far! I left a cosmetic suggestion. Could also upload a small video showcasing the scroll? It'd be nice to see it in action

src/components/Offers/Form/form-components/OfferForm.js
Process-ing: I tried something similar while developing and it does not work well because, for some reason, `errors` and `requestErrors` are both updated when submitting but not at the same time (one first and the other immediately after) and it results in activating the scroll when it shouldn't. For example, when submitting successfully, it would scroll up while transitioning to the new page.

I see that, in your version, you don't check the length of `requestErrors`, so it may not work properly with such errors (like setting the publication date in the past).

src/components/Offers/Form/form-components/OfferForm.js
Process-ing: I needed to place an id for each, but the `MultiOptionTextField` doesn't let me do it as it is. I chose to do this instead of modifying `MultiOptionTextField`, since it is simpler and it doesn't require renaming a bunch of things (more specifically, the component sets an id for each option and has a variable named `id` in the middle of the confusion).

Process-ing: I removed the divs nonetheless.

Process-ing:
Here's the video exemplifying the scroll:
[Screencast from 07-03-2023 15:59:19.webm](https://user-images.githubusercontent.com/42045371/223477548-f6da124c-430b-469c-a447-f42209187d0d.webm)

src/components/Offers/Form/form-components/OfferForm.js
BrunoRosendo: I understand, maybe we could extract the code to an outside function that takes the respective array as a param. What do you think?

BrunoRosendo:
> Here's the video exemplifying the scroll: [Screencast from 07-03-2023 15:59:19.webm](https://user-images.githubusercontent.com/42045371/223477548-f6da124c-430b-469c-a447-f42209187d0d.webm)

Looks great!

src/components/Offers/Form/form-components/OfferForm.js
Process-ing: Yeah, I agree!

BrunoRosendo (approved):

Naapperas (approved):

PR #316, opened 2023-04-24
Use new Location API for Offer Locations

### LINKED ISSUE #152: Use custom service for location picker
Currently, we are using [Geodb Cities API](https://rapidapi.com/wirefreethought/api/geodb-cities/), which is very incomplete. For example, Aveiro is not available.

We should implement our own service to use in the create offer page location picker. We have an available json file containing every city

### LINKED ISSUE #235: Use new Location API
When https://github.com/NIAEFEUP/nijobs-be/pull/211 gets merged, the LocationService in the Front-End should switch from RapidAPI to our own Location API.

Naapperas:
Closes #235 
Closes #152

### DIFF
--- a/src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.js
+++ b/src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.js
@@ -40,6 +40,7 @@ const SearchURLWidget = () => {
     const classes = useSearchURLWidgetStyles();
     const dispatch = useDispatch();
 
+    // TODO: See if this should come from the window browser object or from the router
     const fullURL = window.location.href;
 
     const copyToClipboard = useCallback(() => {
--- a/src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.spec.js
+++ b/src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.spec.js
@@ -23,7 +23,7 @@ describe("Search URL Widget", () => {
 
             const testString = "test-url";
 
-            global.window = Object.create(window);
+            // global.window = Object.create(window);
             Object.defineProperty(window, "location", {
                 value: {
                     href: testString,
@@ -49,7 +49,7 @@ describe("Search URL Widget", () => {
 
             const testString = "test-string";
 
-            global.window = Object.create(window);
+            // global.window = Object.create(window);
             Object.defineProperty(window, "location", {
                 value: {
                     href: testString,
--- a/src/components/Offers/Edit/EditOfferSchema.js
+++ b/src/components/Offers/Edit/EditOfferSchema.js
@@ -59,8 +59,8 @@ export default yup.object().shape({
         .required(HumanValidationReasons.REQUIRED),
     location: yup.string()
         .nullable(true)
-        .matches(/^([a-zA-Z]+([\s-][a-zA-Z])*)+, ([a-zA-Z]+([\s-][a-zA-Z])*)+$/,
-            HumanValidationReasons.LOCATION_FORMAT)
+        // .matches(/^([a-zA-Z]+([\s-][a-zA-Z])*)+, ([a-zA-Z]+([\s-][a-zA-Z])*)+$/,
+        //     HumanValidationReasons.LOCATION_FORMAT)
         .required(HumanValidationReasons.REQUIRED),
     requirements: yup.array()
         .min(...generateValidationRule("requirements", "minLength", HumanValidationReasons.OPTIONS_TOO_SHORT))
--- a/src/components/Offers/Edit/EditOfferSchema.js
+++ b/src/components/Offers/Edit/EditOfferSchema.js
@@ -59,8 +59,6 @@ export default yup.object().shape({
         .required(HumanValidationReasons.REQUIRED),
     location: yup.string()
         .nullable(true)
-        // .matches(/^([a-zA-Z]+([\s-][a-zA-Z])*)+, ([a-zA-Z]+([\s-][a-zA-Z])*)+$/,
-        //     HumanValidationReasons.LOCATION_FORMAT)
         .required(HumanValidationReasons.REQUIRED),
     requirements: yup.array()
         .min(...generateValidationRule("requirements", "minLength", HumanValidationReasons.OPTIONS_TOO_SHORT))
--- a/src/services/locationSearchService.spec.js
+++ b/src/services/locationSearchService.spec.js
@@ -1,5 +1,7 @@
 
 import config from "../config";
+import Constants from "../utils/Constants";
+
 import { searchCities } from "./locationSearchService";
 
 const API_ENDPOINT = `${config.LOCATION_SERVICE_HOSTNAME}/search`;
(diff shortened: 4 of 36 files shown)


Naapperas:
Screenshot of this working with the location service running on my local machine.
image

Naapperas:
The current build failure is due to there being 88 linting warnings

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.js
```diff
@@ -40,6 +40,7 @@ const SearchURLWidget = () => {
     const classes = useSearchURLWidgetStyles();
     const dispatch = useDispatch();
 
+    // TODO: See if this should come from the window browser object or from the router
```
BrunoRosendo, line 43: not for this PR?

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.spec.js
```diff
@@ -23,7 +23,7 @@ describe("Search URL Widget", () => {
 
             const testString = "test-url";
 
-            global.window = Object.create(window);
+            // global.window = Object.create(window);
```
BrunoRosendo, line 26: is this related to this PR?

src/services/locationSearchService.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        expect(res).toEqual(cities);
+    });
+
+    it("Should handle non successful requests", async () => {
+
+        // Simulate request error
+        // any non-2XX status code will be treated as an error so this will do
+        fetch.mockResponseOnce("{}", { status: 500, headers: { "content-type": "application/json" } });
+
+        expect(await searchCities("this value does not matter")).toEqual([]);
```
BrunoRosendo, line 78: Is the reason for ignoring the error (instead of displaying it) simply because the user can manually type the location?

src/utils/UndoableActionsHandlerProvider.js
```diff
@@ -92,10 +92,19 @@ const UndoableActionsHandlerProvider = ({ children }) => {
         if (Object.keys(actions).length && !closeLock) {
             setCloseLock(true);
             initialBeforeUnloadEventListener.current = window.onbeforeunload;
-            window.onbeforeunload =  () => "Some changes might have not taken effect yet...";
+
+            Object.defineProperty(window, "onbeforeunload", {
+                value: () => "Some changes might have not taken effect yet...",
+                writable: true,
```
BrunoRosendo, line 98: why the change?

src/utils/index.js
```diff
@@ -30,13 +30,11 @@ export const capitalize = (str) => {
...
-                return filters[key]
-                    .map((val) => `${key}[]=${encodeURIComponent(val)}`)
-                    .join("&");
-            } else return `${key}=${encodeURIComponent(filters[key])}`;
-        } else return "";
+        if (Array.isArray(filters[key])) {
+            return filters[key]
+                .map((val) => `${key}[]=${encodeURIComponent(val)}`)
+                .join("&");
+        } else return `${key}=${encodeURIComponent(filters[key])}`;
```
BrunoRosendo, line 37: why the change?

src/components/Offers/Edit/EditOfferSchema.js
```diff
@@ -59,8 +59,6 @@ export default yup.object().shape({
         .required(HumanValidationReasons.REQUIRED),
     location: yup.string()
         .nullable(true)
-        .matches(/^([a-zA-Z]+([\s-][a-zA-Z])*)+, ([a-zA-Z]+([\s-][a-zA-Z])*)+$/,
```
BrunoRosendo, line 62: Does the new api not follow the format? Why should we remove validation?

BrunoRosendo (changes requested):
Good job, updating the dev tools is a nice plus. I just left some questions

src/components/Offers/Edit/EditOfferSchema.js
Naapperas: By having the validation here the request never reached the location service, I just removed it so we could have the request. I cannot think of any useful validation other than to check for strings. Any suggestions?

src/utils/index.js
Naapperas: Coverage, since we moved to Codecov's Github app the coverage workflows would prevent this PR from being merged normally. Besides that, the change is technically a no-op since the check that was removed was always true (because of the previous falsy value filtering). It is not related to this PR but was done in order to get the coverage to pass. I believe I have lowered the threshold but it still does not hurt.

src/utils/UndoableActionsHandlerProvider.js
Naapperas: I believe there have been some changes in some of our dependencies that made tests using this bit of code (and other similar ones that are commented as well so I'm gonna mark them as resolved) fail with a runtime error, saying that "onbeforeunload" is not assignable. This fixes that.

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.spec.js
Naapperas: Same thing as the `onbeforeunload` change.

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.js
Naapperas: This PR is just about the location service, but since this could be tackled in the future i just left the TODO there. IDEs like to highlight these so someone may be interested :)

src/services/locationSearchService.spec.js
Naapperas: I'm not sure I understood what you meant. I simply put the service call inside the expect because the service should return `[]` on an API error (mocked by the 500 status code) and no further validation needs to be done, in my opinion. I don't know if that answers your question.

src/components/HomePage/SearchResultsArea/SearchResultsWidget/SearchURLWidget.js
BrunoRosendo: Sure, you might want to track with an issue too

src/services/locationSearchService.spec.js
BrunoRosendo: What I'm saying is this happens:
- User types `Porto`
- API returns error
- Nothing happens and the user types `Porto, Portugal`

instead of
- User types `Porto`
- API returns error
- Connectivty error is shown under the form field and the user types `Porto, Portugal`

That makes sense to me, I'm just confirming

src/components/Offers/Edit/EditOfferSchema.js
BrunoRosendo: Hmm that's weird, before we had the validation and still sent the field to the API, right?

Naapperas: Yes, but we would send the field in the format `city, country` to the old API: that format is not recognized by the new Location API. I can make it so it is recognized but personally i do not believe it should be the case: the API should receive any piece of textual data and return the closest match it finds, without needing a specific format.

Naapperas: I just remembered, whatever is applied to the edit schema should be applied to the creation schema

BrunoRosendo: Fair enough, it makes sense if the API supports all formats

BrunoRosendo (approved):

pedronunes19 (approved):

PR #318, opened 2023-05-20
Port analytics to GA4

### LINKED ISSUE #230: Switch from Universal Analytics to Google Analytics 4
UA is going to be shut down on July 1st 2023. This is a problem because react-ga still uses this version of GA and we'll need to make the switch soon:
https://support.google.com/analytics/answer/10759417?hl=en

BrunoRosendo:
closes #230 

This PR ports the analytics system from GA3/UA to Google Analytics 4.

This should be currently enough to have it running but I still need to test all the functionalities we had previously. I also want to check which new functionalities we might want to use or change from GA3, since GA4 remade how data is handled.

Since GA4 is completely different from GA3, it doesn't seem to be possible to port data over so we'll have to start recording again from the day we release the new version.

I also created new GA4 properties in NIJobs analytics account but I still need to check for extra configurations

### DIFF
--- a/src/utils/analytics/index.js
+++ b/src/utils/analytics/index.js
@@ -1,4 +1,4 @@
-import ReactGa from "react-ga";
+import ReactGA from "react-ga4";
 import { EVENT_TYPES, TIMED_ACTIONS, DIMENSION_IDS, QUERY_VALUE_PARAMETER } from "./constants";
 
 /**
@@ -12,7 +12,7 @@ export const removeCookie = (cookie) => {
  * Initializes Google Analytics.
  */
 export const initAnalytics = () => {
-    ReactGa.initialize(process.env.ANALYTICS_ID, {
+    ReactGA.initialize(process.env.ANALYTICS_ID, {
         gaOptions: {
             siteSpeedSampleRate: 100,
         },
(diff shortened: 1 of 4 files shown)


BrunoRosendo:
Data is already flowing to the new property in GA4. It'll also continue flowing to GA3 until 1st of July.

GA4 has a complete new interface and data flow but Google makes the transition easier by automatically translating old mechanisms to the new ones. GA4 is IMO simpler than GA3, everything is based on events and event parameters.

This means we can still use the old API (e.g. dimensions and old event system) and they'll all be converted into events and event parameters. This also means our code isn't very clear when compared to the new dashboard. I think we can change that later since porting over is what's mostly important right now.

Additionally, since the dashboard is brand new, we should also setup reports and charts in the new properties, which can be done in the staging property after merging to dev. GA4 offers more flexibility and less OOTB reports, so we need to generate our own.

I will be pushing the final version (code barely changes) in a while and asking for review

BrunoRosendo:
I've yet to successfully generate charts using our previous custom dimensions but it should be possible, by the documentation

image

Nevertheless, I created #319 to use the new event system properly

BrunoRosendo:
I also created #320 for the archive

BrunoRosendo:
I found some problems with the events, I need to figure out how to properly set this up to be completely compatible with timings and other events https://www.thyngster.com/performance-timing-tracking-with-google-analytics-4

Naapperas (approved):
LGTM.

Left just a simple question but it does not need to be fixed.

src/utils/analytics/index.js
```diff
@@ -54,10 +54,12 @@ export const measureTime = (action = TIMED_ACTIONS.UNKNOWN, func, label) => asyn
...
-    ReactGa.timing({
-        ...action,
-        value: t1 - t0,
-        label,
+    ReactGA.send({
+        hitType: "timing",
+        timingCategory: action.category,
+        timingVar: action.variable,
+        timingLabel: label,
+        timingValue: t1 - t0,
```
Naapperas, line 62: I recall `react-ga4` having a specialized timing function, but maybe I'm mixing things up.

BrunoRosendo: I don't think so. There isn't any special timing feature in ga4 either, it's just a regular event

BrunoRosendo:
@Naapperas can you do something about codecov? There is nothing to test on those lines

PR #321, opened 2023-06-04
Show indication that an offer is disabled in the "Offers management" page

### LINKED ISSUE #304: Show indication that an offer is disabled in the "Offers management" page
We should show an icon or another type of indicator so companies know when their offers are disabled due to them expiring.

Process-ing:
Closes #304 

At the moment, the code is doing what it should, with the exception that the offer visibility actions are not modifying the chips accordingly (when toggling the visibility, the chips don't change).

From what I have tested, it seems that, for some reason, when the state is modified in the offer visibility actions component, the change only appears inside that component and in none of the rest (like it's just a copy), although I'm using the respective `setState`. I still need to figure out how to fix this.

### DIFF
--- a/src/components/Company/Offers/Manage/CompanyOffersTitle.js
+++ b/src/components/Company/Offers/Manage/CompanyOffersTitle.js
@@ -1,40 +1,43 @@
 import React, { useEffect, useState } from "react";
 import PropTypes from "prop-types";
-import { Chip } from "@material-ui/core";
+import { Chip, makeStyles } from "@material-ui/core";
 
-export const statusChips = {
+const statusChips = {
     hidden: <Chip label="Hidden" size="small" style={{ backgroundColor: "#90A4AE" }} />,
     blocked: <Chip label="Blocked" size="small" style={{ backgroundColor: "#DC4338" }} />,
     archived: <Chip label="Archived" size="small" style={{ backgroundColor: "#56A8D6" }} />,
 };
 
-const OfferTitle = ({ title, offersVisibility, setOfferVisibility, offerId }) => {
+const useStyles = makeStyles(() => ({
+    chips: {
+        position: "absolute",
+    },
+}));
+
+const OfferTitle = ({ title, getOfferVisibility, setOfferVisibility, offerId }) => {
     const [chips, setChips] = useState([]);
-    const isHidden = offersVisibility[offerId]?.isHidden;
-    const isBlocked = offersVisibility[offerId]?.isDisabled;
-    const isArchived = offersVisibility[offerId]?.isArchived;
+    const isHidden = getOfferVisibility(offerId)?.isHidden;
+    const isBlocked = getOfferVisibility(offerId)?.isDisabled;
+    const isArchived = getOfferVisibility(offerId)?.isArchived;
+
+    const classes = useStyles();
 
     useEffect(() => {
         const tempChips = [];
         if (isHidden)
-            tempChips.push(statusChips.hidden);
+            tempChips.push(statusChips.hidden, <span>&nbsp;</span>);
         if (isBlocked)
-            tempChips.push(statusChips.blocked);
+            tempChips.push(statusChips.blocked, <span>&nbsp;</span>);
         if (isArchived)
-            tempChips.push(statusChips.archived);
+            tempChips.push(statusChips.archived, <span>&nbsp;</span>);
         setChips(tempChips);
-    }, [isArchived, isBlocked, isHidden, offerId, offersVisibility, setOfferVisibility]);
+    }, [isArchived, isBlocked, isHidden, setOfferVisibility]);
 
     return (
         <>
             {title}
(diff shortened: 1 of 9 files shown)


src/components/Company/Offers/Manage/CompanyOffersTitle.js
```diff
@@ -0,0 +1,50 @@
...
+    return (
+        <>
+            {title}
+            <div style={{ position: "absolute" }}>
+                {chips.reduce((res, chip, index, chips) => {
+                    res.push(chip);
+                    if (index !== chips.length - 1)
+                        res.push(<span>&nbsp;</span>);
+                    return res;
+                }, [])}
```
Naapperas: Normally this pattern (taking an array of objects and creating an array of JSX elements, which is parsed directly by JSX engines) is done through the use of the `map` function: you map each data element to a UI element. I'll give an example:

```js
function Component({ ... }) {
    ...
    const data = [
        { label: "", url: "" },
        { label: "", url: "" },
    ];
    ...

    return (
        {
            data.map(({ label, url }) => (
                <a href={url}>{label}</a>
            ))
        }
    );
}
```

If I understood your approach, you made use of the `reduce` function to introduce the "non-breaking space" character but I fail to see its use here in favor of, say, styling the elements.

I'm not saying this is wrong, only that I found it strange. I'm leaving this as a change request but please do argue about this.

src/components/Company/Offers/Manage/CompanyOffersTitle.js
```diff
@@ -0,0 +1,50 @@
...
+            tempChips.push(statusChips.blocked);
+        if (isArchived)
+            tempChips.push(statusChips.archived);
+        setChips(tempChips);
+    }, [isArchived, isBlocked, isHidden, offerId, offersVisibility, setOfferVisibility]);
+
+    return (
+        <>
+            {title}
+            <div style={{ position: "absolute" }}>
```
Naapperas: Even though it is not the best solution now (MaterialUI is in version 5, we are using version 4, one of the breaking changes is related to the styling solution which changed completely) you should not have to place styles directly on elements like you did here: why not create a separate style for it?
That way extending the styles for this component can be done in a more centralized way (near other style definitions).

src/components/Company/Offers/Manage/CompanyOffersTitle.js
```diff
@@ -0,0 +1,50 @@
...
+    hidden: <Chip label="Hidden" size="small" style={{ backgroundColor: "#90A4AE" }} />,
+    blocked: <Chip label="Blocked" size="small" style={{ backgroundColor: "#DC4338" }} />,
+    archived: <Chip label="Archived" size="small" style={{ backgroundColor: "#56A8D6" }} />,
+};
+
+const OfferTitle = ({ title, offersVisibility, setOfferVisibility, offerId }) => {
+    const [chips, setChips] = useState([]);
+    const isHidden = offersVisibility[offerId]?.isHidden;
+    const isBlocked = offersVisibility[offerId]?.isDisabled;
+    const isArchived = offersVisibility[offerId]?.isArchived;
```
Naapperas: This code screams "REACT CONTEXT" to me. I know we talked about this in a meeting where we thought it un-wise to use it now because it would mean replacing the entire table implementation we use, but if you do come up with a solution please do let us know.

src/utils/Table/utils.js
```diff
@@ -34,10 +34,12 @@ export const GenerateTableCellFromField = (id, fieldId, fieldOptions, labelId) =
                 id={id === 0 ? `${labelId}-label` : undefined}
                 align={fieldOptions.align || "right"}
             >
-                {fieldOptions?.linkDestination ?
-                    <Link to={linkDestination} className={classes.fieldLink}>
-                        {fieldOptions?.value}
-                    </Link> : fieldOptions.value}
+                <div style={{ marginBlock: extended && "20px" }}>
```
Naapperas: If `extended` is `false` won't this style have a weird value? What happens in that case?

Naapperas (changes requested):
Overall good work.

Left some comments on things that can be improved or just need some arguing over.

src/utils/Table/utils.js
Process-ing: Yes, you're right, it becomes `false`. I should have used a ternary operator. The result is the same nonetheless (the margin becomes zero).

Process-ing:
I finally fixed this issue: the `setState` problem was due to me forgetting to copy the array ;-;

 Now, this PR is working. Here's a video of the final result:
 
[Screencast from 15-07-2023 15:32:44.webm](https://github.com/NIAEFEUP/nijobs-fe/assets/42045371/b6f242f4-147b-4930-b4e8-df473d798eb6)

FranciscoCardoso913 (approved):
Overall the code seems good to me and it works well.

Naapperas (approved):

